Issue #2, opened 2020-10-27
Issue Template


AGlass0fMilk:
We should come up with an issue template, similar to what mbed-os has, which includes required fields for service proposals.

We can follow the lead of BLE standard specifications that can be found here:
https://www.bluetooth.com/specifications/gatt/

Current working format (example with DeviceInformationService):

### Proposed UUID(s):

| Svc Name                 | Char Name         | UUID   | Qualifier | Mandatory Props | Optional Props | Security |
|--------------------------|-------------------|--------|-----------|-----------------|----------------|----------|
| DeviceInformationService | -                 | 0x180A | -         | -               | -              | -        |
| -                        | ModelNumberString | 0x2A24 | O         | Read            |                | None     |

## Brief Description and Intent:

The Device Information Service exposes manufacturer and/or vendor information about a device.

### ModelNumberString Characteristic
The Model Number String characteristic shall represent the model number that is assigned by the device vendor. 

**Characteristic Behavior**: The Model Number String characteristic returns its value when read using the GATT Characteristic Value Read procedure.

pan-:
That's a great start @AGlass0fMilk , I think we need the following things to structure the activity: 

1.  RFC/Service proposal template: I think that's what you're proposing above
1.  Issue template for raising defects 
1.  PR template to propose additions 
1.  Unit test infrastructure (base CMake file with mocks/stubs and utility)
1.  [ADR](https://github.com/joelparkerhenderson/architecture_decision_record) template to store architecture decisions 
1.  Circle CI integration to run unit tests.

Of course the list can be extended but 1, 4, 6 seems to be needed first. What do you think ? 

On the template above, I like the format it foes straight to the point.  I wonder if `Must be compatible with baremetal` should be part of an ADR. It makes reuse of services consistent across build targets.

AGlass0fMilk:
@pan- I agree, I think getting testing setup is a high first priority.

I started with a service proposal template because I wanted to put together a standard format before making issues for a bunch of new services... then we can prioritize what we decide are the most popular services first.

Once we have the infrastructure in place I would like to make a GitHub project to triage the initial batch of proposed services.

Is there a way we can set up automated integration testing? I imagine we would need some BLE targets set up on a publicly-accessible server somewhere. I would host it but my home IP address isn't static :smile: 

[...]

PR #8, opened 2020-10-28
Create template for service proposal 

pan-:


pan-:
@AGlass0fMilk Follow up of #2

.github/ISSUE_TEMPLATE/service-proposal.md
```diff
@@ -0,0 +1,39 @@
...
+
+Use the following template when submitting proposals for new CUSTOM services to this repository.
+
+ Linking to a standard characteristic/service specification document on the Bluetooth SIG's website is sufficient for standard characteristic/service proposals. When proposing standard services, also include specific requirement information for the Mbed BLE implementation of this service (eg: memory constraints, baremetal support, etc) 
+
+The following site can be used to generate a long-format UUID for custom services/characteristics: https://www.uuidgenerator.net/
+This site is useful for generating markdown-formatted tables: https://www.tablesgenerator.com/markdown_tables
+-->
+
+### BLE Standard Service: Yes/No
```
paul-szczepanek-arm, line 20: maybe we can use a tickbox `[ ]` fro that

pan-: That's a good point.

.github/ISSUE_TEMPLATE/service-proposal.md
```diff
@@ -0,0 +1,39 @@
...
+about: Template to use to submit a service proposal
+title: "[SERVICE]"
+labels: service
+assignees: ''
+
+---
+
+<!-- 
+
+Use the following template when submitting proposals for new CUSTOM services to this repository.
```
paul-szczepanek-arm, line 12: so are they custom or can they also be standard services?

.github/ISSUE_TEMPLATE/service-proposal.md
```diff
@@ -0,0 +1,39 @@
...
+ Linking to a standard characteristic/service specification document on the Bluetooth SIG's website is sufficient for standard characteristic/service proposals. When proposing standard services, also include specific requirement information for the Mbed BLE implementation of this service (eg: memory constraints, baremetal support, etc) 
+
+The following site can be used to generate a long-format UUID for custom services/characteristics: https://www.uuidgenerator.net/
+This site is useful for generating markdown-formatted tables: https://www.tablesgenerator.com/markdown_tables
+-->
+
+### BLE Standard Service: Yes/No
+### Proposed UUID(s) or link to specification document:
+
+| Svc Name                 | Char Name         | UUID   | Qualifier | Mandatory Props | Optional Props | Security |
```
paul-szczepanek-arm, line 23: I wouldn't know how to fill out this table. Maybe all this needs is a simple list:
service name - uuid
characteristic name - uuid - mandatory/optional

paul-szczepanek-arm (approved):
I'd expect the implementations would concentrate on standard bluetooth profiles.

.github/ISSUE_TEMPLATE/service-proposal.md
AGlass0fMilk: I'm on the fence about whether the table needs all these fields. I copied the same format that the Bluetooth SIG uses in their service/characteristic specification documents. 

It may be overkill but I can see where each field would be useful. It also helps users think in more detail about how their service operates -- eg: _"what kind of security requirements should this service have?"_, and so on.

A list would probably be easier for users to format as well...

.github/ISSUE_TEMPLATE/service-proposal.md
AGlass0fMilk: They can be standard, but most of the details can be excluded since the official specification document can be linked to instead. 

The more explicit template is for custom services that do not already have a specification document available.

Perhaps we can say: "Use this template to provide details of the proposed service if there is not an existing specification document available."

pan- (commented):
I can't approve my PR but good change, thanks @paul-szczepanek-arm  and @AGlass0fMilk .

AGlass0fMilk:
@paul-szczepanek-arm Please review the changes I made based on our conversation. I have updated the document to clarify when more detailed information must be included in the issue itself.

By "custom" I mean not Bluetooth SIG adopted (ie: it does _not_ have a short-format, 16-bit UUID). If a document already exists that covers the requirements of the service/characteristic proposal template, that document can be attached/linked to instead.

That being said, the user submitting the proposal should still give a brief overview of what the service is, why they think it should be included, and also any foreseeable specific implementation requirements in regards to Mbed.

paul-szczepanek-arm (approved):

PR #10, opened 2020-10-28
AD: Initial test infrastructure

pan-:


doc/adr/0002-define-initial-testing-infrastructure.md
```diff
@@ -0,0 +1,62 @@
...
+We will use `google test` to unit test services proposed. The tests will be compiled 
+with `Clang` and will generate a code coverage report when run.
+For pull request validation, the code and test suite will be compiled and run by `Circle CI`. 
+Depending on results and code coverage, the submission will be rejected. 
+
+## Consequences
+
+The base of the test infrastructure should be created, it must be based on google test, 
+be compiled with clang run in Circle CI. 
+Maintainers should not merge PR with less than 80% code coverage.
```
paul-szczepanek-arm, line 62: who are those maintainers you speak of?

paul-szczepanek-arm (approved):

paul-szczepanek-arm:
IMHO this is a bit too heavy for what this is meant to be. Will this attract contributors or appear as a barrier to them?

doc/adr/0002-define-initial-testing-infrastructure.md
pan-: Us for now, selected members of the community latter.

pan-:
@paul-szczepanek-arm Thanks for the feedback Paul, that's a proposition and certainly the point that requires the most debate. I put 80% as an arbitrary measure but it can be dropped to lower value or even put has a recommendation. Maybe it should not even be part of this ADR which would just focus on the initial test infrastructure (gtest + mock + circle CI).

AGlass0fMilk:
> IMHO this is a bit too heavy for what this is meant to be. Will this attract contributors or appear as a barrier to them?

The comprehensive testing and CI that many of Mbed's official repositories have is one of their most comforting aspects. My previous bad experiences with the old, untested Mbed BLE services is a motivating factor behind why this repository was created.

The services and extensions here should be high quality, and that quality must be ensured through rigorous testing. I think once the example is set it won't be too hard to leverage existing tests when creating new ones.

That being said, I don't have any idea what level of work is required to attain 80% code coverage :sweat_smile: 

So perhaps we make that a suggested value here and determine a concrete minimum later on once we see what is realistic?

pan-:
@paul-szczepanek-arm @AGlass0fMilk I updated the PR based on your feedback and removed code coverage requirement, this can be the subject of another AD when we have a better understanding of what would be a reasonable threshold and the amount of work required to get there.

AGlass0fMilk:
LGTM

AGlass0fMilk (approved):

PR #11, opened 2020-10-28
Test base infrastructure

pan-:
This PR provides stubs of mbed BLE API and necessary infrastructure to build tests. 

It requires approvable of #10 .


pan-:
@paul-szczepanek-arm Can you add support to build this in Circle CI ?

paul-szczepanek-arm (approved):
would be useful if the dummy service had a characteristic to show how it's tested

AGlass0fMilk (commented):
Where are these stubs from? Were they preexisting or did you make them for this project?

Some of the files seem to be from mbed-os. If so, it would be nice to have a script or some automated way to copy them over (if it is not possible to pull them in as a dependency from mbed-os itself).

I'm sure there's a reason you did it this way, I just want to understand why. It seems not ideal from a maintenance point of view.

I suppose we will have to update this repository periodically when mbed-os releases are made. Similar to the BLE examples, right?

pan-:
@AGlass0fMilk I extracted headers I deemed relevant from Mbed OS. There is just two actual stubs: one for the GattServer and another for the BLE instance.  ([here](https://github.com/ARMmbed/mbed-os-experimental-ble-services/tree/11efaa905aaf2cac8b46cb15b4f1c0ae9a9f2232/tests/stubs/source)). Both are very simple. 

I'm sure other stubs will be required over time, for example, a bond management service would need access to the SecurityManager. 

I understand your concern on maintenance and they are shared because this can get out of sync very quickly. However I'm fairly certain the API of the GattServer will remain stable for the next six month at least. We have no immediate plan on reworking the GattServer API; If there are changes they will be driven by work made in this repository.

My proposal would be to go with what is in this PR to start then adjust depending on the overhead/issues syncing with Mbed OS induce. Note that stub code needs to be modified to expose virtual functions instead of non virtual member functions. I'm sure it can be automated with libclang or some python script. Or even better, provided by Mbed OS!  

I should document the process in this PR though 😅 .

PR #14, opened 2020-11-26
Add Link Loss Service

noonfom:
The Link Loss Service uses the Alert Level characteristic, as defined in the [Bluetooth SIG Assigned Numbers](https://www.bluetooth.com/specifications/assigned-numbers/), to cause an
alert in the device when the link is lost.

# Reviewers
@pan- 
@paul-szczepanek-arm

### DIFF
--- a/services/LinkLoss/source/LinkLossService.cpp
+++ b/services/LinkLoss/source/LinkLossService.cpp
@@ -1,0 +1,110 @@
+/* mbed Microcontroller Library
+ * Copyright (c) 2020 ARM Limited
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#if BLE_FEATURE_GATT_SERVER
+
+#include "ble-service-link-loss/LinkLossService.h"
+
+LinkLossService::LinkLossService(BLE &ble, events::EventQueue &event_queue, ChainableGapEventHandler &chainable_gap_event_handler) :
+    _ble(ble),
+    _event_queue(event_queue),
+    _chainable_gap_event_handler(chainable_gap_event_handler),
+    _alert_level_char(GattCharacteristic::UUID_ALERT_LEVEL_CHAR, &_alert_level)
+{
+}
+
+LinkLossService::~LinkLossService()
+{
+    _event_queue.cancel(_event_queue_handle);
+}
+
+void LinkLossService::init()
+{
+    GattCharacteristic *charTable[] = { &_alert_level_char };
+    GattService         linkLossService(GattService::UUID_LINK_LOSS_SERVICE, charTable, 1);
+
+    _alert_level_char.setWriteAuthorizationCallback(this, &LinkLossService::onDataWritten);
+
+    ble_error_t error = _ble.gattServer().addService(linkLossService);
+
+    if (error == BLE_ERROR_NONE) {
+        _chainable_gap_event_handler.addEventHandler(this);
+    }
+}
+
+void LinkLossService::set_event_handler(EventHandler* handler)
+{
+    _alert_handler = handler;
+}
+
+void LinkLossService::set_alert_level(AlertLevel level)
+{
+    _alert_level = level;
+}
+
+void LinkLossService::set_alert_timeout(std::chrono::milliseconds timeout)
+{
+    _alert_timeout = timeout;
+}
+
+LinkLossService::AlertLevel LinkLossService::get_alert_level()
+{
+    return _alert_level;
+}
+
+void LinkLossService::stop_alert()
+{
+    _event_queue.cancel(_event_queue_handle);
+    _event_queue_handle = 0;
+    if (_in_alert) {
+        _in_alert = false;
+        if (_alert_handler) {
+            _alert_handler->on_alert_end();
+        }
+    }
+}
+
+void LinkLossService::onConnectionComplete(const ble::ConnectionCompleteEvent &event)
+{
+    if (event.getStatus() == BLE_ERROR_NONE) {
+        stop_alert();
+    }
+}
+
+void LinkLossService::onDisconnectionComplete(const ble::DisconnectionCompleteEvent &event)
+{
+    AlertLevel level = get_alert_level();
+    if (_alert_handler != nullptr && event.getReason() == ble::disconnection_reason_t::CONNECTION_TIMEOUT &&
+        level != AlertLevel::NO_ALERT && !_in_alert) {
+        _in_alert = true;
+        _alert_handler->on_alert_requested(level);
+        if (_alert_timeout > std::chrono::milliseconds(0)) {
+            _event_queue_handle = _event_queue.call_in(_alert_timeout, [this] { stop_alert(); });
+        }
+    }
+}
+
+void LinkLossService::onDataWritten(GattWriteAuthCallbackParams *write_request)
+{
+    const uint8_t level = *write_request->data;
+
+    if (level <= (uint8_t)(AlertLevel::HIGH_ALERT)) {
+        set_alert_level((AlertLevel) level);
+    }
+}
+
+#endif // BLE_FEATURE_GATT_SERVER
--- a/services/LinkLoss/source/LinkLossService.cpp
+++ b/services/LinkLoss/source/LinkLossService.cpp
@@ -104,6 +104,9 @@ void LinkLossService::onDataWritten(GattWriteAuthCallbackParams *write_request)
 
     if (level <= (uint8_t)(AlertLevel::HIGH_ALERT)) {
         set_alert_level((AlertLevel) level);
+    } else {
+        // The alert level is out of range
+        write_request->authorizationReply = static_cast<GattAuthCallbackReply_t>(0xFF);
     }
 }
 
--- a/services/LinkLoss/include/ble-service-link-loss/LinkLossService.h
+++ b/services/LinkLoss/include/ble-service-link-loss/LinkLossService.h
@@ -1,0 +1,168 @@
+/* mbed Microcontroller Library
+ * Copyright (c) 2020 ARM Limited
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef LINK_LOSS_SERVICE_H
+#define LINK_LOSS_SERVICE_H
+
+#if BLE_FEATURE_GATT_SERVER
+
+#include "ble/BLE.h"
+#include "ble/Gap.h"
+#include "events/EventQueue.h"
+#include "ble/gap/ChainableGapEventHandler.h"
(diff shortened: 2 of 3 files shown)


services/src/LinkLossService/LinkLossService.cpp
```diff
@@ -0,0 +1,110 @@
...
+        in_alert = false;
+        if (alert_handler) {
+            alert_handler->on_alert_end();
+        }
+    }
+}
+
+void LinkLossService::onConnectionComplete(const ble::ConnectionCompleteEvent &event)
+{
+    stop_alert();
```
paul-szczepanek-arm, line 81: you need to check the status of the event

services/src/LinkLossService/LinkLossService.cpp
```diff
@@ -0,0 +1,110 @@
...
+LinkLossService::LinkLossService(BLE &ble, events::EventQueue &event_queue, ChainableGapEventHandler &chainable_gap_event_handler) :
+    _ble(ble),
+    _event_queue(event_queue),
+    _chainable_gap_event_handler(chainable_gap_event_handler),
+    _alert_level_char(GattCharacteristic::UUID_ALERT_LEVEL_CHAR, &_alert_level)
+{
+}
+
+LinkLossService::~LinkLossService()
+{
```
paul-szczepanek-arm, line 29: you need to cancel the pending timeout:
_event_queue.cancel(_event_queue_handle);

services/src/LinkLossService/LinkLossService.cpp
```diff
@@ -0,0 +1,110 @@
...
+    if (write_request->offset != 0) {
+        write_request->authorizationReply = AUTH_CALLBACK_REPLY_ATTERR_INVALID_OFFSET;
+        return;
+    }
+
+    if (level <= *reinterpret_cast<uint8_t *>(AlertLevel::HIGH_ALERT)) {
+        set_level((AlertLevel) level);
+    } else {
+        write_request->authorizationReply = AUTH_CALLBACK_REPLY_ATTERR_WRITE_NOT_PERMITTED;
+    }
```
paul-szczepanek-arm, line 109: if the BLE device is compliant it cannot write anything else in the service, maybe we can dispense with the checking (you'd still need to check the value before the enum cast)

paul-szczepanek-arm (changes requested):
Good work, please fill out the doxy.

paul-szczepanek-arm:
We need to decide on some common folder structure for the service files. @pan- @AGlass0fMilk ?

services/inc/LinkLossService.h
```diff
@@ -0,0 +1,126 @@
...
+     * @param ble
+     * @param event_queue
+     * @param chainable_gap_event_handler
+     */
+    LinkLossService(BLE &ble, events::EventQueue &event_queue, ChainableGapEventHandler &chainable_gap_event_handler);
+
+    /**
+     *
+     */
+    ~LinkLossService();
```
pan-, line 67: This class is not copyable or movable, please delete the copy constructor and copy assignment operator: 

```c++
LinkLossService(const LinkLossService&) = delete;
LinkLossService &operator=(const LinkLossService&) = delete;
```

services/inc/LinkLossService.h
```diff
@@ -0,0 +1,126 @@
...
+    void onDataWritten(GattWriteAuthCallbackParams *write_request);
+
+    BLE &_ble;
+    events::EventQueue &_event_queue;
+    ChainableGapEventHandler &_chainable_gap_event_handler;
+
+    ReadWriteGattCharacteristic<AlertLevel> _alert_level_char;
+    AlertLevel _alert_level;
+    std::chrono::milliseconds _alert_timeout;
+    EventHandler *alert_handler;
```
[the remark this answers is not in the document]
pan-: Missing `_` prefix

services/inc/LinkLossService.h
```diff
@@ -0,0 +1,126 @@
...
+
+    bool in_alert = false;
```
pan-, line 120: missing `_` prefix.

services/src/LinkLossService/LinkLossService.cpp
```diff
@@ -0,0 +1,110 @@
...
+{
+    stop_alert();
+}
+
+void LinkLossService::onDisconnectionComplete(const ble::DisconnectionCompleteEvent &event)
+{
+    AlertLevel level = get_level();
+    if (event.getReason() == ble::disconnection_reason_t::CONNECTION_TIMEOUT && level != AlertLevel::NO_ALERT) {
+        in_alert = true;
+        alert_handler->on_alert_requested(level);
```
pan-, line 89: You should check if `alert_handler` is not a nullptr.

services/src/LinkLossService/LinkLossService.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+void LinkLossService::onDataWritten(GattWriteAuthCallbackParams *write_request)
+{
+    const uint8_t level = *write_request->data;
+
+    if (write_request->offset != 0) {
+        write_request->authorizationReply = AUTH_CALLBACK_REPLY_ATTERR_INVALID_OFFSET;
+        return;
+    }
+
```
pan-, line 104: Please check the size of the write request too.

services/inc/LinkLossService.h
```diff
@@ -0,0 +1,126 @@
...
+public:
+    enum class AlertLevel : uint8_t {
+        NO_ALERT    = 0,
+        MILD_ALERT  = 1,
+        HIGH_ALERT  = 2
+    };
+
+    struct EventHandler {
+        virtual void on_alert_requested(AlertLevel) { }
+        virtual void on_alert_end() { }
```
pan-, line 52: It might be useful to raise an event when the alert level change; for example if you want to store it. What do you thing @paul-szczepanek-arm ?

pan- (changes requested):
Thanks for the submission @noonfom . 

One thing: as a general rule of thumb consider doing in class initialization for class variables that do not depend on the constructor, in this case `_event_queue_handle`, `_alert_timeout`, ...

services/inc/LinkLossService.h
```diff
@@ -0,0 +1,126 @@
...
+#include "ble/gap/ChainableGapEventHandler.h"
+
+#include <chrono>
+
+/**
+ * BLE Link Loss Service
+ *
+ * @par purpose
+ *
+ * @par usage
```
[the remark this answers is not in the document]
noonfom: Purpose will give a basic overview of the service in line with the [LLS_v1.0.1](https://www.bluetooth.org/DocMan/handlers/DownloadDoc.ashx?doc_id=303198)
It might also be useful to add some insight into the local `EventHandler`. The user must override `onAlertRequested()` and `onAlertEnd()` to effectively interact with the user. 
What do you think?

services/inc/LinkLossService.h
```diff
@@ -0,0 +1,126 @@
...
+     *
+     * @param level
+     */
+    void set_level(AlertLevel level);
+
+    /**
+     *
+     * @param timeout
+     */
+    void set_timeout(std::chrono::milliseconds timeout);
```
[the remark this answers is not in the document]
noonfom: I agree. Thanks Paul.

services/src/LinkLossService/LinkLossService.cpp
paul-szczepanek-arm: is all this checking really necessary? doesn't gatt server already check this?

paul-szczepanek-arm: also see my comment below

services/inc/LinkLossService.h
paul-szczepanek-arm: I think that's overkill, this state only makes sense in the context of a "sessions", there is no need to raise an alert after a reset, you've just reset you should be checking if the reset was intended or not and raising "alerts" based on that instead

then again the cost of another handler is only a few bytes

services/inc/LinkLossService.h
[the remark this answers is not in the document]
paul-szczepanek-arm: yes, my advice is to keep it short and link to the spec for more details

services/inc/LinkLossService.h
pan-: I was thinking of the use case where you want to retain the value of the alert in non volatile memory. It is still possible to store it at disconnection though.

services/src/LinkLossService/LinkLossService.cpp
pan-: You are right, the characteristic has a fixed length and a size of 1. If the offset and length are invalid then it should be rejected by the server before the invocation of this callback.

services/inc/LinkLossService.h
```diff
@@ -0,0 +1,126 @@
...
+    events::EventQueue &_event_queue;
+    ChainableGapEventHandler &_chainable_gap_event_handler;
+
+    ReadWriteGattCharacteristic<AlertLevel> _alert_level_char;
+    AlertLevel _alert_level;
+    std::chrono::milliseconds _alert_timeout;
+    EventHandler *alert_handler;
+
+    bool in_alert = false;
+    int _event_queue_handle;
```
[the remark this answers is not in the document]
noonfom: Thanks @paul-szczepanek-arm

services/inc/LinkLossService.h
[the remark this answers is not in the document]
noonfom: Thanks @paul-szczepanek-arm and @pan-

services/src/LinkLossService/LinkLossService.cpp
```diff
@@ -0,0 +1,104 @@
...
+}
+
+LinkLossService::AlertLevel LinkLossService::get_alert_level()
+{
+    return _alert_level;
+}
+
+void LinkLossService::stop_alert()
+{
+    _event_queue.cancel(_event_queue_handle);
```
pan-, line 69: Can you reset `_event_queue_handle` to its default value: `0` when it is cancelled ?

pan- (changes requested):
We're almost there, we just need to cleanup the value of `_event_queue_handle` when it is cancelled (or triggered).

services/src/LinkLossService/LinkLossService.cpp
noonfom: Yes, thanks @pan-

services/src/LinkLossService/LinkLossService.cpp
```diff
@@ -0,0 +1,105 @@
...
+{
+    if (event.getStatus() == BLE_ERROR_NONE) {
+        stop_alert();
+    }
+}
+
+void LinkLossService::onDisconnectionComplete(const ble::DisconnectionCompleteEvent &event)
+{
+    AlertLevel level = get_alert_level();
+    if (_alert_handler != nullptr && event.getReason() == ble::disconnection_reason_t::CONNECTION_TIMEOUT && level != AlertLevel::NO_ALERT) {
```
paul-szczepanek-arm, line 89:
```suggestion
    if (_alert_handler != nullptr && event.getReason() == ble::disconnection_reason_t::CONNECTION_TIMEOUT && level != AlertLevel::NO_ALERT && !_in_alert) {
```

services/src/LinkLossService/LinkLossService.cpp
```diff
@@ -0,0 +1,105 @@
...
+            _event_queue_handle = _event_queue.call_in(_alert_timeout, [this] { stop_alert(); });
+        }
+    }
+}
+
+void LinkLossService::onDataWritten(GattWriteAuthCallbackParams *write_request)
+{
+    const uint8_t level = *write_request->data;
+
+    if (level <= *reinterpret_cast<uint8_t *>(AlertLevel::HIGH_ALERT)) {
```
paul-szczepanek-arm, line 102:
```suggestion
    if (level <= static_cast<uint8_t>(AlertLevel::HIGH_ALERT)) {
```

paul-szczepanek-arm (changes requested):
Just a couple more issues and please add an mbed_lib.json, see:
https://github.com/ARMmbed/mbed-os-experimental-ble-services/pull/15

paul-szczepanek-arm (approved):

services/LinkLoss/source/LinkLossService.cpp
```diff
@@ -0,0 +1,110 @@
...
+    }
+}
+
+void LinkLossService::onDataWritten(GattWriteAuthCallbackParams *write_request)
+{
+    const uint8_t level = *write_request->data;
+
+    if (level <= (uint8_t)(AlertLevel::HIGH_ALERT)) {
+        set_alert_level((AlertLevel) level);
+    }
```
pan-, line 107: Could you return an error if the alert is out of range ?

pan- (changes requested):
Minor change request when invalid data are written from a peer. 
After this it should be good to go.

pan- (approved):

PR #15, opened 2020-12-01
Directory structure adr

paul-szczepanek-arm:
We need to agree on this so we don't clash. Please comment on the proposal.

### DIFF
--- a/doc/adr/0003-directory-structure.md
+++ b/doc/adr/0003-directory-structure.md
@@ -8,7 +8,7 @@ Accepted
 
 ## Context
 
-The will be many services contributed by multiple people. We need a common way of structuring the services
+There will be many services contributed by multiple people. We need a common way of structuring the services
 so it's easy for the user to pick up any one service.
 
 ## Decision
--- a/doc/adr/0003-directory-structure.md
+++ b/doc/adr/0003-directory-structure.md
@@ -11,6 +11,9 @@ Accepted
 There will be many services contributed by multiple people. We need a common way of structuring the services
 so it's easy for the user to pick up any one service.
 
+Some services might want to share some common code. We want to have that code structured so that include path shouldn't
+change if these extensions are included into Mbed OS.
+
 ## Decision
 
 Each service will be its own library. All services will be in the services directory.
@@ -31,6 +34,35 @@ and the mbed_lib.json would contain at least:
     "name": "ble-service-example"
 }
 ```
+
+If a service needs to share some common code between multiple services this code can be made into an extension.
+
+Extensions are common code, extending the mbed-os ble feature. They are libraries of their own and each service that
+wants to use them needs a `requires` section that pulls it in as a dependency.
+
+And `example` extension library:
+
+```
+extensions/example/include/ble/gatt
+extensions/example/include/ble/gap
+extensions/example/include/ble/common
+extensions/example/source/...
+extensions/example/mbed_lib.json
+```
+
+and the extension mbed_lib.json would contain at least:
+
+```
+{
+    "name": "ble-extension-example"
+}
+```
+
+and the service needs it's mbed_lib.json expanded with a
+
+```
+    requires: ["ble-extension-example"]
+``` 
 
 This way a user can clone the repo and include an individual service without getting unwanted includes.
 
--- a/doc/adr/0003-directory-structure.md
+++ b/doc/adr/0003-directory-structure.md
@@ -19,9 +19,9 @@ Each service will contain at least an include folder and a mbed_lib.json.
 An example service called `example` would look like this:
 
 ```
-services/ble-service-example/include/example.h
-services/ble-service-example/source/example.cpp
-services/ble-service-example/mbed_lib.json
+services/example/include/ble-service-example/example.h
+services/example/source/example.cpp
+services/example/mbed_lib.json
 ```
 
 and the mbed_lib.json would contain at least:
--- a/doc/adr/0003-directory-structure.md
+++ b/doc/adr/0003-directory-structure.md
@@ -1,0 +1,39 @@
+# 3. Define initial testing infrastructure
+
+Date: 2020-12-01
+
+## Status
+
+Accepted
+
+## Context
+
+The will be many services contributed by multiple people. We need a common way of structuring the services
+so it's easy for the user to pick up any one service.
+
+## Decision
+
+Each service will be its own library. All services will be in the services directory.
+Each service will contain at least an include folder and a mbed_lib.json.
+
+An example service called `example` would look like this:
+
+```
+services/ble-service-example/include/example.h
+services/ble-service-example/source/example.cpp
+services/ble-service-example/mbed_lib.json
+```
+
+and the mbed_lib.json would contain at least:
+
+```
+{
+    "name": "ble-service-example"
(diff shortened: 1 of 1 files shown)


doc/adr/0003-directory-structure.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Decision
+
+Each service will be its own library. All services will be in the services directory.
+Each service will contain at least an include folder and a mbed_lib.json.
+
+An example service called `example` would look like this:
+
+```
+services/ble-service-example/include/example.h
```
pan-, line 22: That's not in line with what's in mbed-os at the moment which is `include/library-name` to include the header files with `library-name/header.h`.

doc/adr/0003-directory-structure.md
```diff
@@ -0,0 +1,39 @@
...
+services/ble-service-example/source/example.cpp
+services/ble-service-example/mbed_lib.json
+```
+
+and the mbed_lib.json would contain at least:
+
+```
+{
+    "name": "ble-service-example"
```
pan-, line 31: That's great, thanks for thinking of this.

doc/adr/0003-directory-structure.md
```diff
@@ -0,0 +1,39 @@
+# 3. Define initial testing infrastructure
```
pan-, line 1: Please update the name of this ADR.

pan- (commented):
Thanks Paul, what do you think of making the inclusion pattern consistent with Mbed OS ?

pan-:
You forgot to change the include structure in https://github.com/ARMmbed/mbed-os-experimental-ble-services/pull/15/commits/003fe26ca7f7360e78a3ad2d42437d0db718441b

paul-szczepanek-arm:
Sorry, it's the constant timeouts yesterday (it messes up github desktop).

doc/adr/0003-directory-structure.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Decision
+
+Each service will be its own library. All services will be in the services directory.
+Each service will contain at least an include folder and a mbed_lib.json.
+
+An example service called `example` would look like this:
+
+```
+services/example/include/ble-service-example/example.h
```
pan-, line 22: if the example service is named example, shouldn't it be `services/example/include/example/example.h` ?

paul-szczepanek-arm: The prefix is to avoid clashes with non ble services. The library name is called ble-service-example. The directory name for the include is the same as library name.

pan-:
@AGlass0fMilk Are you happy with the proposed structure ?

AGlass0fMilk (changes requested):

paul-szczepanek-arm:
I thought the idea was that once the service stops being experimental it can be moved to mbed-os. Is that not the case?

pan-:
> I thought the idea was that once the service stops being experimental it can be moved to mbed-os. Is that not the case?

It is all about the level of support and integration. Given the trend I think it would not be moved to Mbed OS. 

@AGlass0fMilk It would be great to have reusable utilities/helper classes shared across the services of the repo. 
I can open an ADR to discuss of the location.

doc/adr/0003-directory-structure.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+There will be many services contributed by multiple people. We need a common way of structuring the services
+so it's easy for the user to pick up any one service.
+
```
pan-, line 13: Could you provide a context for user extensions ?

doc/adr/0003-directory-structure.md
```diff
@@ -0,0 +1,39 @@
...
+## Decision
+
```
pan-, line 15: I think we can split the `Decision` in several parts: `services` and `extensions`. Maybe we could mention test that goes into `tests`. I don't think there is a need to detail it at the moment.  

For extensions, what do you think of having them in an `extensions` folder organized like Mbed OS BLE ? 

- `extensions/include/ble/gatt`
- `extensions/include/ble/gap`
- `extensions/include/ble/common`
- `extensions/source/...`

That way, include path shouldn't change if these extensions are included into Mbed OS.

pan- (commented):
Forget about my last comment, this is the ADR for the directory structure. 

I added a proposal for extensions structure. Let me know what you think.

doc/adr/0003-directory-structure.md
paul-szczepanek-arm: sounds good, might as well put this into this adr

doc/adr/0003-directory-structure.md
paul-szczepanek-arm: added, please rereview

pan- (approved):
Thanks @paul-szczepanek-arm it is good to have more details.

pan-:
@AGlass0fMilk Are you satisfied with the updates made to the document ? 
It would be good to reach consensus rapidly so it doesn't block other PRs.

doc/adr/0003-directory-structure.md
```diff
@@ -0,0 +1,88 @@
...
+and the mbed_lib.json would contain at least:
+
+```
+{
+    "name": "ble-service-example"
+}
+```
+
+This way a user can clone the repo and include an individual service without getting unwanted includes.
+
```
AGlass0fMilk, line 41: Not to make this PR take longer to integrate, but what about Client code for a given service?

I'm thinking it should share the same root directory so that headers can be shared between the Service and Client implementations.

For example, I can foresee a `CurrentTimeService` and a corresponding `CurrentTimeServiceClient` (or should it just be `CurrentTimeClient`?)

It would be nice to be able to share definitions (such as UUIDs, etc) while having a separate `mbed_lib.json`. The include and source code for the Client and Service can share the same directories, but I'm wondering what you guys think.

The Client specification should be placed between the Service and Extensions section.

AGlass0fMilk (changes requested):

doc/adr/0003-directory-structure.md
AGlass0fMilk: Perhaps we could treat clients similar to extensions and put them in a separate directory. The client can then "require" the inclusion of the corresponding service.

Does this sound like a reasonable compromise? This would prevent having to pull in client code when the user only wants the service itself, while also allowing the client to share code with the service.

paul-szczepanek-arm: Maybe client code could just be put together with the service? there are #defines that include or exclude code based on whether gatt server or gatt client is present.

paul-szczepanek-arm: But your solution sounds good to me too (with clients being their own lib requiring the service to get the headers). @pan-?

pan-: I like both of your approach; only clients would require the compilation of the server code. Paul's approach has merit too. 

Maybe we can leave it open for now and define a new ADR when the case occur.

AGlass0fMilk: That sounds alright to me. We can address this when the client issue comes up.

AGlass0fMilk (approved):

PR #17, opened 2020-12-09
Add Current Time Service

noonfom:
The current time service (CTS) allows a BLE device to expose date and time information to other BLE devices.

**Specification:** [CTS_SPEC_V1.1.0](https://www.bluetooth.org/docman/handlers/downloaddoc.ashx?doc_id=292957) 

##  Progress
A CTS instance can expose the following characteristics:
- [x] Current Time (M)
- [ ] Local Time Information (O)
- [ ] Reference Time Information (O)

## Issues
None
   
## Useful links
* Older Mbed-OS implementation of the CTS: https://os.mbed.com/users/ohneta/code/BLE_CurrentTimeService/
 
## Reviewers
@pan- 
@paul-szczepanek-arm


services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -0,0 +1,178 @@
...
+        struct tm *current_time;
+
+        current_time->tm_year = (*ct | (*(ct + 1) << 8)) - 1900;
+        ct += 2;
+        current_time->tm_mon  =  *ct++ - 1;
+        current_time->tm_mday =  *ct++;
+        current_time->tm_hour =  *ct++;
+        current_time->tm_min  =  *ct++;
+        current_time->tm_sec  =  *ct++;
+        current_time->tm_wday =  (*ct == 7 ? 0 : *ct);
```
paul-szczepanek-arm, line 160: leave a comment explaining the monday->sunday thing

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -0,0 +1,178 @@
...
+     * Construct and initialize a current time service.
+     *
+     * @param ble
+     * @param event_queue
+     * @param chainable_gap_event_handler
+     */
+    CurrentTimeService(BLE &ble, events::EventQueue &event_queue) :
+            _ble(ble),
+            _event_queue(event_queue),
+            _current_time_char(GattCharacteristic::UUID_CURRENT_TIME_CHAR, &_current_time) {
```
paul-szczepanek-arm, line 89: this is going to work on any sane platform but if you're not using an array of bytes there should be an MBED_ASSERT that check the sizeof _current_time is 10

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -0,0 +1,178 @@
...
+    void set_current_time() {
+        timeval tv{};
+        gettimeofday(&tv, NULL);
+        // printf("%u\r\n", (unsigned int) tv.tv_sec);
+
+        struct tm *current_time;
+        time_t local_time = tv.tv_sec + _current_time_offset;
+        current_time = localtime(&local_time);
+
+        _current_time.year    = current_time->tm_year + 1900;
```
paul-szczepanek-arm, line 127: this will work on little endian targets, to be portable you should ensure the year is stored little endian regardless or host endianness - the gatt server will treat the characteristic as a stream of bytes and will not perform a swap if needed, all it can see is a bunch of bytes that it will send over ATT protocol

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -0,0 +1,178 @@
...
+                                reinterpret_cast<const uint8_t *>(&_current_time),
+                                sizeof(_current_time));
+
+        read_request->authorizationReply = AUTH_CALLBACK_REPLY_SUCCESS;
+    }
+
+    void onDataWritten(GattWriteAuthCallbackParams *write_request) {
+        const uint8_t *ct = (write_request->data);
+
+        struct tm *current_time;
```
paul-szczepanek-arm, line 151:
```suggestion
        struct tm *remote_time_of_day;
```

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -0,0 +1,178 @@
...
+        current_time->tm_year = (*ct | (*(ct + 1) << 8)) - 1900;
+        ct += 2;
+        current_time->tm_mon  =  *ct++ - 1;
+        current_time->tm_mday =  *ct++;
+        current_time->tm_hour =  *ct++;
+        current_time->tm_min  =  *ct++;
+        current_time->tm_sec  =  *ct++;
+        current_time->tm_wday =  (*ct == 7 ? 0 : *ct);
+
+        time_t remote_time = mktime(current_time);
```
paul-szczepanek-arm, line 162:
```suggestion
        time_t remote_time_in_seconds = mktime(current_time);
```

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -0,0 +1,178 @@
...
+        uint8_t  day;
+        uint8_t  hours;
+        uint8_t  minutes;
+        uint8_t  seconds;
+
+        uint8_t  weekday;
+
+        uint8_t  fractions256;
+
+        uint8_t  adjustReason;
```
paul-szczepanek-arm, line 56: some comments about the values would help understand the contents next time someone has to edit this
```suggestion
        uint16_t year; // calendar year based at year 1, with valid values 1582-9999
        uint8_t  month; // 1-12
        uint8_t  day; //calendar day
        uint8_t  hours;
        uint8_t  minutes;
        uint8_t  seconds;

        uint8_t  weekday; // 1-7 starting with Monday

        uint8_t  fractions256;

        uint8_t  adjustReason;
```

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -0,0 +1,178 @@
...
+    };
+
+    struct EventHandler {
+        /**
+         * On current time changed
+         *
+         * This function is called if the current time characteristic is changed by the client
+         *
+         * @attention This is an abstract function and should be overridden by the user.
```
paul-szczepanek-arm, line 65: this function is not abstract, it has an no op implementation

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -0,0 +1,178 @@
...
+        virtual void on_current_time_written(time_t current_time) {}
+    };
+
+    /**
+     * Set event handler
+     *
+     * @param handler EventHandler object to handle events raised by the current time service
+     */
+    void set_event_handler(EventHandler *handler) {
+
```
paul-szczepanek-arm, line 76: missing

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -0,0 +1,178 @@
...
+    time_t get_current_time() {
+        time_t t = time(NULL);
+
+        return _current_time_offset + t;
+    }
+
+    void set_time(time_t new_time) {
+        time_t t = time(NULL);
+
+        _current_time_offset = t - new_time;
```
paul-szczepanek-arm, line 114: imagine what would happen if the current system time was 1000 and the user called set_time(1010). The user has set the time in the future but the offset is now negative (1000 - 1010 = -10). So when he gets time with get_current_time() he will get: 1000 + (-10) = 990 which is in the past.

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+private:
+    void set_current_time() {
+        timeval tv{};
+        gettimeofday(&tv, NULL);
+        // printf("%u\r\n", (unsigned int) tv.tv_sec);
+
+        struct tm *current_time;
```
paul-szczepanek-arm, line 123: you can declare it on the same line as you use it to make it more readable

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -0,0 +1,178 @@
...
+        time_t local_time = tv.tv_sec + _current_time_offset;
+        current_time = localtime(&local_time);
```
paul-szczepanek-arm, line 125: the variable names are all a bit close, how about something more verbose
```suggestion
        new_time_of_day = localtime(&new_time_in_seconds);
```

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    void set_time(time_t new_time) {
+        time_t t = time(NULL);
+
+        _current_time_offset = t - new_time;
+    }
+
+private:
+    void set_current_time() {
```
paul-szczepanek-arm, line 118: set implies you give it the value, maybe update?
```suggestion
    void update_current_time() {
```

paul-szczepanek-arm (changes requested):
Thanks. You'll need to remove linkloss from this PR and add an mbed_lib.json for the current time service.

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -0,0 +1,178 @@
...
+     */
+    CurrentTimeService(BLE &ble, events::EventQueue &event_queue) :
+            _ble(ble),
+            _event_queue(event_queue),
+            _current_time_char(GattCharacteristic::UUID_CURRENT_TIME_CHAR, &_current_time) {
+        memset(&_current_time, 0, sizeof(_current_time));
+    }
+
+    ble_error_t init() {
+        GattCharacteristic *charTable[] = {&_current_time_char};
```
paul-szczepanek-arm, line 94: this is not specific to this service, but we must find a better way of creating services, we're allocating memory for a characteristic only to have the gatt server allocating more memory and then we keep that characteristic even though the only thing we care about is the handle the gatt server sets on it when it creates the service

there must be sane way to crate a service without this waste

@pan- does the gatt server depends on the existence of the characteristic in any way after the service is added?

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
pan-: @paul-szczepanek-arm We provided default implementations for every handler so far. Maybe we should stay consistent. If new handlers are added, it won't require to change `= 0` into `{}`. I don't have strong opinion on this.

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
pan-: That's a good point Paul, I think it is safe to create the characteristic on the stack. What we are really interrested in are the handles assigned by the server. 
I need to double check.

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -0,0 +1,178 @@
...
+     *
+     * @param ble
+     * @param event_queue
+     * @param chainable_gap_event_handler
+     */
+    CurrentTimeService(BLE &ble, events::EventQueue &event_queue) :
+            _ble(ble),
+            _event_queue(event_queue),
+            _current_time_char(GattCharacteristic::UUID_CURRENT_TIME_CHAR, &_current_time) {
+        memset(&_current_time, 0, sizeof(_current_time));
```
pan-, line 90: Initialization can be solved at declaration: `CurrentTime _current_time = { 0 };`

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
pan-: Endianness utilities are missing in Mbed OS. Probably something we could add in the common set of tools for this repo.

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -0,0 +1,178 @@
...
+        _current_time.seconds = current_time->tm_sec;
+        _current_time.weekday = (current_time->tm_wday == 0) ? 7 : current_time->tm_wday;
+        // 1/256th of seconds not supported
+        _current_time.fractions256 = 0;
+    }
+
+    void onDataRead(GattReadAuthCallbackParams *read_request) {
+        set_current_time();
+
+        _ble.gattServer().write(_current_time_char.getValueHandle(),
```
pan-, line 141: I'm not sure it is necessary to write the characteristic. The request expose a `data` pointer and `len` field that can be changed to point to `_current_time`.

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -0,0 +1,178 @@
...
+        current_time->tm_mon  =  *ct++ - 1;
+        current_time->tm_mday =  *ct++;
+        current_time->tm_hour =  *ct++;
+        current_time->tm_min  =  *ct++;
+        current_time->tm_sec  =  *ct++;
+        current_time->tm_wday =  (*ct == 7 ? 0 : *ct);
+
+        time_t remote_time = mktime(current_time);
+
+        set_time(remote_time);
```
pan-, line 164: Doesn't this require data validation for all the fields ?

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    BLE &_ble;
+    events::EventQueue &_event_queue;
+
+    CurrentTime _current_time;
```
pan-, line 170: Two options to make sure the structure as a proper layout: 
- Use `MBED_PACKED`
- Wrap this into an union with an array of byte of the size wanted then static assert the size of the union.

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
paul-szczepanek-arm: I wasn't suggesting we make this abstract. I think default implementation is good because saves the user from having to define it if he only needs one event (OK, this one so far has only one event so this point is moot). I was just pointing out the comment contradicts the code.

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
paul-szczepanek-arm: personally I think the spec guarantees the data is correct and checks are only needed if data can cause an error or a security issue but in this case worst can happen is the time is set to nonsense

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
noonfom: I agree but perhaps `local_time_tm` and `remote_time_tm` are better names since the struct holds more than just the time of day?

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
noonfom: Isn't the default time in seconds? For example, wouldn't it be redundant to write `epoch_time_in_seconds = time(NULL)`

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
noonfom: Thanks Paul. The comment is a misuse of the term "abstract function". This mistake is also present in LLS, so I'll correct it there too.

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -0,0 +1,236 @@
...
+    ble_error_t init() {
+        GattCharacteristic *charTable[] = {&_current_time_char};
+        GattService currentTimeService(GattService::UUID_CURRENT_TIME_SERVICE, charTable, 1);
+
+        _current_time_char.setReadAuthorizationCallback (this, &CurrentTimeService::onDataRead);
+        _current_time_char.setWriteAuthorizationCallback(this, &CurrentTimeService::onDataWritten);
+
+        ble_error_t bleError = _ble.gattServer().addService(currentTimeService);
+
+        MBED_ASSERT(sizeof(_current_time) == CURRENT_TIME_CHAR_VALUE_SIZE);
```
paul-szczepanek-arm, line 121: this can probably be a static assert since it's known at compile time

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -42,19 +42,55 @@ ble_error_t CurrentTimeService::init()
...
     _time_offset = new_time - epoch_time;
 }
 
-void CurrentTimeService::onDataRead(GattReadAuthCallbackParams *read_request) {
+bool CurrentTimeService::current_time_is_valid()
+{
+    bool result = true;
+
+    if ((_current_time.year    < 1582) || (_current_time.year    > 9999)) {
+        result = false;
```
paul-szczepanek-arm, line 68: don't continue checking all the other if conditions if you already know it's wrong, just return false here and now

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -42,19 +42,55 @@ ble_error_t CurrentTimeService::init()
...
+    }
+    if ((_current_time.month   <    1) || (_current_time.month   >   12)) {
+        result = false;
+    }
+    if ((_current_time.day     <    1) || (_current_time.day     >   31)) {
+        result = false;
+    }
+    if ((_current_time.hours   <    0) || (_current_time.hours   >   23)) {
```
paul-szczepanek-arm, line 76: it's unsigned so not possible for it to be less than 0

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -68,12 +104,19 @@ void CurrentTimeService::onDataRead(GattReadAuthCallbackParams *read_request) {
...
 
     struct tm remote_time_tm{};
 
-    deserialize(&remote_time_tm, data);
+    write_request->authorizationReply = AUTH_CALLBACK_REPLY_SUCCESS;
+
+    if (!deserialize(&remote_time_tm, data)) {
+        /* The alert level is out of range
+         * */
+        write_request->authorizationReply = static_cast<GattAuthCallbackReply_t>(0xFF);
```
paul-szczepanek-arm, line 118: you send an error but the code continues past the if statement

how about:
```
if (deserialise) {
  call handler
} else {
  send error
}
```

paul-szczepanek-arm: also you don't need ```static_cast<GattAuthCallbackReply_t>(0xFF)``` any more, Vincent added the missing AUTH_CALLBACK_REPLY_ATTERR_OUT_OF_RANGE error and others and they've been merged to mbedos

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -99,25 +141,46 @@ void CurrentTimeService::serialize(uint8_t *data, const struct tm *local_time_tm
...
+         * So, if weekday = 7, i.e. Sunday, the correct value for tm_wday is 0
+         * Otherwise, the fields signify the same days and no correction is needed
+         * */
+        remote_time_tm->tm_wday  = (_current_time.weekday == 7 ? 0 : _current_time.weekday);
+
+        deserialization_is_possible = true;
+    } else {
+        /* Invalid data written to CurrentTime struct, can't deserialize
+         * */
+        _current_time = { 0 };
```
paul-szczepanek-arm, line 181: no need to zero it, it's never read outside of here

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -42,19 +42,55 @@ ble_error_t CurrentTimeService::init()
...
-void CurrentTimeService::set_time(time_t new_time) {
+void CurrentTimeService::set_time(time_t new_time)
+{
     time_t epoch_time = time(nullptr);
 
     _time_offset = new_time - epoch_time;
 }
 
-void CurrentTimeService::onDataRead(GattReadAuthCallbackParams *read_request) {
+bool CurrentTimeService::current_time_is_valid()
```
paul-szczepanek-arm, line 63: that's a great way of doing it, this way you can run the same check on send!

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -68,12 +104,19 @@ void CurrentTimeService::onDataRead(GattReadAuthCallbackParams *read_request) {
     read_request->authorizationReply = AUTH_CALLBACK_REPLY_SUCCESS;
```
paul-szczepanek-arm, line 104: before you set the data and the AUTH_CALLBACK_REPLY_SUCCESS check `current_time_is_valid()` and if it's not send an error instead

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -0,0 +1,186 @@
...
+     * */
+    *data++ =  local_time_tm->tm_wday == 0 ? 7 : local_time_tm ->tm_wday;
+    *data   =  0;
+}
+
+bool CurrentTimeService::deserialize(struct tm *remote_time_tm, const uint8_t *data)
+{
+    bool deserialization_is_possible;
+
+    _current_time.year    = *data | (*(data + 1) << 8);
```
pan-, line 153: I don't expect deserialization to mutate members of the class. It means than in `onDataWritten`, it can fail and return an error but the data are committed in the service. 

I'd suggest changing `current_time_is_valid` into a member function of `CurrentTime` (`CurrentTime::is_valid`). Then add a new deserialize function from `uint8_t*` to `CurrentTime` and a conversion function from `CurrentTime` to `tm` (`struct tm toTm()`) in the `CurrentTime` class.

paul-szczepanek-arm: I think it's fine because the _current_time member is not really used for anything - it's like a scratch space, the real time value is stored in the _time_offset which is only updated on successful deserialisation.

pan-: It is **not** ok, the function `current_time_is_valid` is public, serialization of rejected value should not have a visible side effect to the application.

paul-szczepanek-arm: current_time_is_valid shouldn't be public, it's not actually validating 'current' time, it's only validating what has been sent

paul-szczepanek-arm: that member is really just scratch space, not really backing anything, the use cannot read it and the gatt server doesn't  use it

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
pan-: Answer it is not. I close but that is something we should consider as an improvement for Q1.

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -43,6 +44,13 @@
  */
 class CurrentTimeService {
 public:
+    mbed::Timer timer;
+
+    const uint8_t MANUAL_TIME_UPDATE             = 1 << 0;
```
pan-, line 49: please mark it static so it doesn't take space into an instance and stays in flash.

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -43,6 +44,13 @@
  */
 class CurrentTimeService {
 public:
+    mbed::Timer timer;
```
pan-, line 47: This is of no interest for the user and we don't them to mess with the timer. This should be a private member.

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -92,15 +100,20 @@ class CurrentTimeService {
      * Set the time offset, i.e. the time in seconds beyond Epoch time.
      *
      * @param host_time Time in seconds according to your host.
-     *
+     * @param adjust_reason
```
pan-, line 103: Please provide documentation for the parameter.

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -45,18 +50,60 @@ void CurrentTimeService::set_event_handler(EventHandler *handler) {
...
+
+        timer.reset();
+        timer.start();
+    }
+
+    start_periodic_time_update();
+}
+
+void CurrentTimeService::start_periodic_time_update() {
+    if (_event_queue_handle == 0) {
```
pan-, line 101: Can you point me out in the spec when it is required to do a periodic update ?

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -45,18 +50,60 @@ void CurrentTimeService::set_event_handler(EventHandler *handler) {
...
+    time_t local_time = get_time();
+
+    struct tm *local_time_tm = localtime(&local_time);
+
+    CurrentTime current_time(local_time_tm);
+
+    current_time.adjust_reason = adjust_reason;
+
+    bool send_update = true;
+    if ((current_time.adjust_reason & (MANUAL_TIME_UPDATE | CHANGE_OF_TIME_ZONE | CHANGE_OF_DST)) == 0) {
```
pan-, line 80: I don't think that is correct, `(EXTERNAL_REFERENCE_TIME_UPDATE | CHANGE_OF_TIME_ZONE)` should trigger an update.

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -0,0 +1,236 @@
...
+void CurrentTimeService::onCurrentTimeWritten(GattWriteAuthCallbackParams *write_request)
+{
+    CurrentTime input_time(write_request->data);
+
+    if (!input_time.valid()) {
+        write_request->authorizationReply = AUTH_CALLBACK_REPLY_ATTERR_OUT_OF_RANGE;
+        return;
+    }
+
+    if (input_time.fractions256 || input_time.adjust_reason) {
```
paul-szczepanek-arm, line 133: I don't understand why the ```|| input_time.adjust_reason``` is there, can you explain in a comment please?

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -45,18 +50,60 @@ void CurrentTimeService::set_event_handler(EventHandler *handler) {
...
+    struct tm *local_time_tm = localtime(&local_time);
+
+    CurrentTime current_time(local_time_tm);
+
+    current_time.adjust_reason = adjust_reason;
+
+    bool send_update = true;
+    if ((current_time.adjust_reason & (MANUAL_TIME_UPDATE | CHANGE_OF_TIME_ZONE | CHANGE_OF_DST)) == 0) {
+        if (timer.elapsed_time() < 15min) {
+            if (std::abs(time_offset_difference) <= 1) {
```
pan-, line 82: `time_t` is in second so it should be `60` instead of `1`.

pan- (commented):
Thanks for the submission, I'd like to understand where in the spec it is indicated that the server should push an update every 15 minutes.

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -0,0 +1,236 @@
...
+    CurrentTime input_time(write_request->data);
+
+    if (!input_time.valid()) {
+        write_request->authorizationReply = AUTH_CALLBACK_REPLY_ATTERR_OUT_OF_RANGE;
+        return;
+    }
+
+    if (input_time.fractions256 || input_time.adjust_reason) {
+        write_request->authorizationReply = (GattAuthCallbackReply_t) DATA_FIELD_IGNORED;
+        return;
```
paul-szczepanek-arm, line 135: why is this returning? you can ignore the fractions field and send back an error but according to the spec the time should still be adjusted

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    struct tm remote_time_tm{};
+    input_time.to_tm(&remote_time_tm);
+    time_t remote_time = mktime(&remote_time_tm);
+
+    set_time(remote_time, MANUAL_TIME_UPDATE);
```
paul-szczepanek-arm, line 142: why are you not using the `input_time.adjust_reason` as the reason?

paul-szczepanek-arm (changes requested):
still reviewing

services/CurrentTime/source/CurrentTimeService.cpp
paul-szczepanek-arm: By negation of:
"If the time of the server device is changed because of reference time update, then no notifications shall be sent to the client device within the 15 minutes from the last notification"

Negating this: notification will be sent after 15 minutes.

services/CurrentTime/source/CurrentTimeService.cpp
pan-: We don't know how to react for external `adjust_reason` yet. The specification doesn't help.

services/CurrentTime/source/CurrentTimeService.cpp
pan-: Interesting, I didn't interpreted it that way. I was thinking the notification would not be sent at all.

pan-: But in such case, the periodic time update should not be periodic and only queued if `send_update ` was false in `update_current_time_value`. It also means that when an update is sent, the pending event should be removed. 

What happens to the reason too ? It doesn't seem to be properly captured by `start_periodic_time_update` and if an update requiring immediate notification is sent while another one is pending, shouldn't it add `EXTERNAL_REFERENCE_TIME_UPDATE` to the reasons ?

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -0,0 +1,236 @@
...
+{
+    time_t epoch_time = time(nullptr);
+
+    time_t old_time_offset = _time_offset;
+    _time_offset = host_time - epoch_time;
+
+    update_current_time_value(adjust_reason, old_time_offset - _time_offset);
+}
+
+void CurrentTimeService::update_current_time_value(const uint8_t adjust_reason, time_t time_offset_difference) {
```
pan-, line 70: Could you change `adjust_reason` into `adjust_reasons` ? This byte may contain several reasons. Please apply the same change where relevant.

services/CurrentTime/source/CurrentTimeService.cpp
paul-szczepanek-arm: CHANGE_OF_TIME_ZONE already triggers an update (default value of send_update is true) and EXTERNAL_REFERENCE_TIME_UPDATE will also send an update but only if 15 minutes have elapsed (as mandated by spec)

services/CurrentTime/source/CurrentTimeService.cpp
noonfom: I added that before `adjust_reason` was used in `set_time()`. There's no reason to ignore that field anymore. What do you think @pan- ?

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -0,0 +1,236 @@
...
+        }
+    }
+
+    if (send_update) {
+        _ble.gattServer().write(_current_time_char.getValueHandle(),
+                                reinterpret_cast<const uint8_t *>(&current_time),
+                                CURRENT_TIME_CHAR_VALUE_SIZE);
+
+        timer.reset();
+        timer.start();
```
paul-szczepanek-arm, line 94: we should cancel and zero the _event_queue_handle here

services/CurrentTime/source/CurrentTimeService.cpp
paul-szczepanek-arm: technically the EXTERNAL_REFERENCE_TIME_UPDATE reason should be set every second and cleared when notification is sent

services/CurrentTime/source/CurrentTimeService.cpp
paul-szczepanek-arm: I think since the adjust reason is trasmitted in the payload it should be written as is - the spec does not mention it, so there's not reason to assume it has any different meaning

paul-szczepanek-arm: The client must follow the spec same as servers, so they apply the same rules to the adjust reason they send

services/CurrentTime/source/CurrentTimeService.cpp
noonfom: How do you suggest resolving this? I guess by updating the `authorizationReply` after calling `on_current_time_changed` ?

services/CurrentTime/source/CurrentTimeService.cpp
noonfom: I guess but the field is not given a plural name in the spec. The value can signify a compound of reasons without having an 's' at the end.

services/CurrentTime/source/CurrentTimeService.cpp
paul-szczepanek-arm: of course the above does not apply to user interaction - so if a user sets it with set time or a remote peer sends a value then whatever reason they have given should be the reason used

paul-szczepanek-arm: since in effect the only time a EXTERNAL_REFERENCE_TIME_UPDATE is sent is because of the periodic update, it's OK to just rely on the periodic update function to give the reason as otherwise the reason is already supplied correctly

services/CurrentTime/source/CurrentTimeService.cpp
paul-szczepanek-arm: just remove the `return`, you send the error that *some* fields were ignored but you still set the new time

paul-szczepanek-arm:
Please don't force push when you make fixes.

services/CurrentTime/source/CurrentTimeService.cpp
pan-: Damn, I'm an idiot. Thanks @paul-szczepanek-arm .

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -0,0 +1,186 @@
...
+    static const uint8_t MANUAL_TIME_UPDATE             = 1 << 0;
+    static const uint8_t EXTERNAL_REFERENCE_TIME_UPDATE = 1 << 1;
+    static const uint8_t CHANGE_OF_TIME_ZONE            = 1 << 2;
+    static const uint8_t CHANGE_OF_DST                  = 1 << 3;
+
+    static constexpr std::chrono::seconds UPDATE_TIME_PERIOD = std::chrono::seconds(60);
+
+    struct EventHandler {
+        /**
+         * This function is called if the current time characteristic is changed by the client
```
paul-szczepanek-arm, line 57:
```suggestion
         * This function is called if the current time characteristic is changed by the client
         * @param current_time Newly set current time in seconds
         * @param adjust_reason Reason for the change, a bitmask of any one of MANUAL_TIME_UPDATE, EXTERNAL_REFERENCE_TIME_UPDATE, CHANGE_OF_TIME_ZONE, CHANGE_OF_DST or a combination of thereof.
```

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -0,0 +1,227 @@
...
+    }
+}
+
+void CurrentTimeService::onCurrentTimeRead(GattReadAuthCallbackParams *read_request)
+{
+    time_t local_time = get_time();
+    CurrentTime local_current_time(localtime(&local_time));
+
+    if (local_current_time.valid()) {
+        _current_time = local_current_time;
```
paul-szczepanek-arm, line 107: there's no point to assign this

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -0,0 +1,186 @@
...
+     *
+     * @return Time in seconds.
+     */
+    time_t get_time() const;
+
+    /**
+     * Set the time offset, i.e. the time in seconds beyond Epoch time.
+     *
+     * @param host_time Time in seconds according to your host.
+     * @param adjust_reason Bitmask using a combination of MANUAL_TIME_UPDATE, EXTERNAL_REFERENCE_TIME_UPDATE,
```
paul-szczepanek-arm, line 103:
```suggestion
     * @param adjust_reason Bitmask using one or a combination of MANUAL_TIME_UPDATE, EXTERNAL_REFERENCE_TIME_UPDATE,
```

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -0,0 +1,227 @@
...
+    if (input_time.fractions256) {
+        write_request->authorizationReply = (GattAuthCallbackReply_t) DATA_FIELD_IGNORED;
+    } else {
+        write_request->authorizationReply = AUTH_CALLBACK_REPLY_SUCCESS;
+    }
+}
+
+CurrentTimeService::CurrentTime::CurrentTime(const uint8_t *data)
+{
+    // FIXME: 2bit transmitted in little endian
```
paul-szczepanek-arm, line 144: ?

noonfom: I'm not sure about this. Can you please elaborate on this comment @pan- ?

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
noonfom: A combination can exclude all but one?

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -0,0 +1,227 @@
...
+    year          = *data | (*(data + 1) << 8);
```
paul-szczepanek-arm, line 145: the data is already in the correct order

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -0,0 +1,227 @@
...
+    minutes       = *data++;
+    seconds       = *data++;
+    weekday       = *data++;
+    fractions256  = *data++;
+    adjust_reason = *data;
+}
+
+CurrentTimeService::CurrentTime::CurrentTime(const struct tm *local_time_tm)
+{
+    // FIXME: tm_year should be coded in little endian
```
paul-szczepanek-arm, line 158: yes, please fix

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -0,0 +1,227 @@
...
+     * */
+    weekday       =  local_time_tm->tm_wday == 0 ? 7 : local_time_tm ->tm_wday;
+    fractions256  =  0;
+    adjust_reason =  0;
+}
+
+
+bool CurrentTimeService::CurrentTime::valid()
+{
+    if ((year    < 1582) || (year    > 9999)) {
```
paul-szczepanek-arm, line 179: you cannot read the year directly, it needs to be deserialised

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -0,0 +1,227 @@
...
+    return true;
+}
+
+bool CurrentTimeService::CurrentTime::to_tm(struct tm * remote_time_tm)
+{
+    if (!valid()) {
+        return false;
+    }
+
+    remote_time_tm->tm_year  =  year  - 1900;
```
paul-szczepanek-arm, line 210: you cannot read the year directly, it needs to be deserialised

services/CurrentTime/source/CurrentTimeService.cpp
paul-szczepanek-arm: I suggest making an accessor for year

services/CurrentTime/source/CurrentTimeService.cpp
noonfom: It's deserialised in the constructor

services/CurrentTime/source/CurrentTimeService.cpp
paul-szczepanek-arm: I was referring to your own code comment,
you can leave the comment just remove the `FIXME:` it can remain as a informative comment

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
paul-szczepanek-arm: good point, technically you can have an unknown reason
```suggestion
     * @param adjust_reason Bitmask of one of or a combination of MANUAL_TIME_UPDATE, EXTERNAL_REFERENCE_TIME_UPDATE or 0 if reason is unknown,
```

services/CurrentTime/source/CurrentTimeService.cpp
paul-szczepanek-arm: it should not be

paul-szczepanek-arm: oh, wait, you're not sending this out

noonfom: I'm not sending it out. The data is passed in by the client and a pointer to the data steam is used to construct a `CurrentTime` object in `onCurrentTimeRead()` The constructor deserialises the data. This function simply converts a `CurrentTime` object to libc `tm` struct.

paul-szczepanek-arm: line 108

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -125,6 +125,13 @@ class CurrentTimeService {
 
         bool to_tm(struct tm * remote_time_tm);
 
+        uint16_t get_year() {
+            uint8_t *year_data  = (uint8_t *)&year;
+            uint16_t year_value = *year_data;
+            year_value |= *year_data << 8;
+            return year_value;
```
pan-, line 132: You can just return `year` for now and leave a comment that the code should be modified when BE support is added.

services/CurrentTime/source/CurrentTimeService.cpp
paul-szczepanek-arm: this is not resolved, this whole thing could just be a memcpy

services/CurrentTime/source/CurrentTimeService.cpp
noonfom: This thread began with a comment about the `to_tm` member which is called in on data written.

services/CurrentTime/source/CurrentTimeService.cpp
pan-: I think it is important to keep it in that form so ww know what to change if we add BE support. 
Not a blocker for me at the moment.

paul-szczepanek-arm: OK, but when we do have BE support then it will be a memcpy - which is my point but yeah, not a blocker

pan-: Yes, we can optimize it when that support is added. That is internal details today and a small micro optimization. It is fine to keep it like it today.

services/CurrentTime/source/CurrentTimeService.cpp
paul-szczepanek-arm: Let me explain this:
you have a struct that has multibyte values
you also use the same struct a buffer which you then send verbatim over the air
the value needs to be stored therefore in network order
at the same time you read that value in your code on your host, this means every time you read this value stored in network order you need to convert it to host order
in most cases on arm host is little endian same as network order so this is a nop
on a big endian this will break

services/CurrentTime/include/ble-service-current-time/CurrentTimeService.h
```diff
@@ -101,7 +101,7 @@ class CurrentTimeService {
      *
      * @param host_time Time in seconds according to your host.
      * @param adjust_reason Bitmask using a combination of MANUAL_TIME_UPDATE, EXTERNAL_REFERENCE_TIME_UPDATE,
-     * CHANGE_OF_TIME_ZONE and CHANGE_OF_DST representing the reason for setting the time.
+     * CHANGE_OF_TIME_ZONE and CHANGE_OF_DST representing the reason for setting the time or zero if year is unknown.
```
paul-szczepanek-arm, line 104: reason not year

paul-szczepanek-arm (changes requested):
fix comment and ship it

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -0,0 +1,226 @@
...
+        write_request->authorizationReply = (GattAuthCallbackReply_t) DATA_FIELD_IGNORED;
+    } else {
+        write_request->authorizationReply = AUTH_CALLBACK_REPLY_SUCCESS;
+    }
+}
+
+CurrentTimeService::CurrentTime::CurrentTime(const uint8_t *data)
+{
+    year          = *data | (*(data + 1) << 8);
+    month         = *data++;
```
paul-szczepanek-arm, line 146: wrong byte, you need to increment by 2

paul-szczepanek-arm, line 146: data += 2; missing

services/CurrentTime/source/CurrentTimeService.cpp
```diff
@@ -0,0 +1,231 @@
...
+        read_request->authorizationReply = AUTH_CALLBACK_REPLY_ATTERR_UNLIKELY_ERROR;
+    }
+}
+
+void CurrentTimeService::onCurrentTimeWritten(GattWriteAuthCallbackParams *write_request)
+{
+    CurrentTime input_time(write_request->data);
+
+    if (write_request->len != CURRENT_TIME_CHAR_VALUE_SIZE) {
+        write_request->authorizationReply = AUTH_CALLBACK_REPLY_ATTERR_INVALID_ATTRIBUTE_VALUE_LENGTH;
```
paul-szczepanek-arm, line 122: missing return

paul-szczepanek-arm (approved):

PR #18, opened 2020-12-17
Implement Gatt Presentation Format Descriptor class

AGlass0fMilk:
This commit introduces a class that simplifies the creation of a Gatt Presentation Format Descriptor. This descriptor is a Bluetooth SIG-specified descriptor type with the short-form UUID `0x2904`.

See full official specification here: https://www.bluetooth.com/wp-content/uploads/Sitecore-Media-Library/Gatt/Xml/Descriptors/org.bluetooth.descriptor.gatt.characteristic_presentation_format.xml

### DIFF
--- a/descriptors/GattPresentationFormatDescriptor.h
+++ b/descriptors/GattPresentationFormatDescriptor.h
@@ -1,0 +1,60 @@
+/* mbed Microcontroller Library
+ * Copyright (c) 2020-2021 Embedded Planet, Inc
+ * Copyright (c) 2020-2021 ARM Limited
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef GATT_PRESENTATION_FORMAT_DESCRIPTOR_H_
+#define GATT_PRESENTATION_FORMAT_DESCRIPTOR_H_
+
+#include "GattCharacteristic.h"
+#include "GattAttribute.h"
+
+/** Length of a presentation format descriptor struct */
+#define GATT_PRESENTATION_FORMAT_DESCRIPTOR_LEN 7
+
+/**
+ * Class encapsulating a GATT Presentation Format Descriptor
+ */
+class GattPresentationFormatDescriptor: public GattAttribute {
+
+public:
+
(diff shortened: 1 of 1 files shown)


AGlass0fMilk:
Migrating this from [ep-oc-mcu](https://github.com/EmbeddedPlanet/ep-oc-mcu/blob/master/features/FEATURE_BLE/GattPresentationFormatDescriptor.h)

descriptors/GattPresentationFormatDescriptor.h
```diff
@@ -0,0 +1,66 @@
...
+			memcpy(&format[5], &namespace_description, sizeof(namespace_description));
+
+		}
+
+	private:
+
+		// Wouldn't it be nice if packing structs was more consistently supported by compilers?
+		//struct GattCharacteristic::PresentationFormat_t format;
+
+		// In lieu of using the struct above, packing issues makes us have to use a raw buffer
```
paul-szczepanek-arm, line 62: we have
```MBED_PACKED(struct)```

File: descriptors/GattPresentationFormatDescriptor.h [hunk 2]
```diff
@@ -0,0 +1,66 @@
...
+			/** Populate the format struct */
+//			format.gatt_format = format_type;
+//			format.exponent = exponent;
+//			format.gatt_unit = unit;
+//			format.gatt_namespace = namespace_id;
+//			format.gatt_nsdesc = namespace_description;
+
+			format[0] = format_type;
+			format[1] = exponent;
+			memcpy(&format[2], &unit, sizeof(unit));
```
paul-szczepanek-arm, line 51: needs some hton function, we really should put one in BLE in mbed-os, same on line 53

paul-szczepanek-arm (changes requested):
Tabs - > spaces.
You can have a packed struct if you want and a static assert that assures the packed size is correct.
We really need some accessible (de)serialisation utils in mbed's BLE

pan-:
> We really need some accessible (de)serialisation utils in mbed's BLE

Better if it was accessible to Mbed OS, we're even missing little/big endian extraction. I like the Cordio one https://github.com/ARMmbed/mbed-os/blob/master/connectivity/FEATURE_BLE/libraries/cordio_stack/wsf/include/util/bstream.h

AGlass0fMilk:
> > We really need some accessible (de)serialisation utils in mbed's BLE
> 
> Better if it was accessible to Mbed OS, we're even missing little/big endian extraction. I like the Cordio one https://github.com/ARMmbed/mbed-os/blob/master/connectivity/FEATURE_BLE/libraries/cordio_stack/wsf/include/util/bstream.h

Would like to have a class/C++ based implementation too.

descriptors/GattPresentationFormatDescriptor.h
AGlass0fMilk: I think it's more straightforward to just use a raw byte buffer.

descriptors/GattPresentationFormatDescriptor.h
AGlass0fMilk: I implemented the endianness conversion using bitwise math.

AGlass0fMilk:
Took care of the tabs/spaces conversion and cleaned up the dead commented code.

PR #23, opened 2021-01-21
requires is not needed

paul-szczepanek-arm:


AGlass0fMilk:
Is this true for the new tools?

What about in baremetal mode? Perhaps it should be mentioned if you must do this in baremetal mode.

paul-szczepanek-arm:
The new tools ignores it, the old tools ignores it (which was a surprise to me). We don't support baremetal yet.

pan-:
> We don't support baremetal yet.
I'm certain Bluetooth supports baremetal. 

Maybe you can just indicates the require step is for baremetal applications.

paul-szczepanek-arm:
OK, in which case it's a different kind of requires, let me fix that.

pan- (approved):

PR #27, opened 2021-02-02
Add sequence diagrams

noonfom:
This PR adds UML sequence diagrams for the Link Loss Service.


pan-:
Thanks for the submission @noonfom , these seems to be between sequence and activity diagram.

plantuml/LinkLoss/connection.puml
```diff
@@ -0,0 +1,24 @@
...
+
+|Gap|
+start
+
+|Link Loss Service|
+:connection;
+-> event;
+
+if (event.getStatus() == BLE_ERROR_NONE ?) then (yes)
+    :stop_alert()|
```
paul-szczepanek-arm, line 16: shouldn't this check if alert is actually happening first?

plantuml/LinkLoss/data_written.puml
```diff
@@ -0,0 +1,28 @@
...
+
+|Link Loss Service|
+:data written;
+-> write_request;
+
+:set alert level based on write request;
+-> level;
+
+if (level <= HIGH_ALERT ?) then (yes)
+    :set_alert_level(level)|
```
paul-szczepanek-arm, line 19: I don't get it, who is "APP" in this context? do you mean service or user? I don't think we notify the user

plantuml/LinkLoss/disconnection.puml
```diff
@@ -0,0 +1,46 @@
...
+-> level;
+
+if (alert_handler != NULL ?) then (yes)
+    if (event.getReason() == CONNECTION_TIMEOUT ?) then (yes)
+        if (level != NO_ALERT ? ) then (yes)
+            if (~ in alert ?) then (yes)
+                :set alert flag to true;
+                :on_alert_requested()|
+                if (alert_timeout > 0 ms ?) then (yes)
+                    :event_queue.call_in(alert_timeout, stop_alert())|
```
paul-szczepanek-arm, line 25: not sure we need such detail, including event queue

plantuml/LinkLoss/disconnection.puml
```diff
@@ -0,0 +1,46 @@
...
+
+:get_alert_level()|
+-> level;
+
+if (alert_handler != NULL ?) then (yes)
+    if (event.getReason() == CONNECTION_TIMEOUT ?) then (yes)
+        if (level != NO_ALERT ? ) then (yes)
+            if (~ in alert ?) then (yes)
+                :set alert flag to true;
+                :on_alert_requested()|
```
paul-szczepanek-arm, line 23: this should be in the app column

plantuml/LinkLoss/init.puml
```diff
@@ -0,0 +1,28 @@
...
+|Link Loss Service|
+:init;
+
+:construct gatt service for lls;
+
+:set write auth. callback for alert_level_char.;
+
+|Gatt Server|
+:add lls to gatt server;
+-> error;
```
paul-szczepanek-arm, line 16: so what happens on error?
maybe init should return ble_error_t

paul-szczepanek-arm (changes requested):

paul-szczepanek-arm:
that's just my opinion but I think diagrams like that should be more human readable and not paste actual code

plantuml/LinkLoss/connection.puml
noonfom: That check is performed in `stop_alert`. Please look at the sequence diagram for the function: [stop_alert.png](https://github.com/noonfom/mbed-os-experimental-ble-services/blob/seq-diag/plantuml/LinkLoss/img/stop_alert.png) or read the PlantUML: [stop_alert.puml](https://github.com/noonfom/mbed-os-experimental-ble-services/blob/seq-diag/plantuml/LinkLoss/stop_alert.puml). This is __not__ present in the implementation file: [LinkLossService.cpp](https://github.com/noonfom/mbed-os-experimental-ble-services/blob/ebf327d801610bb3261edd0521f43b7cdf0bc459/services/LinkLoss/source/LinkLossService.cpp#L81).

plantuml/LinkLoss/data_written.puml
noonfom: In this context, "App" refers to the user application, i.e. the program that instantiates a Link Loss Service.

I am not entirely sure what you mean by: "I don't think we notify the user". If you are referring to call to `set_alert_level`, please see line 106 of the implementation file: [LinkLossService](https://github.com/noonfom/mbed-os-experimental-ble-services/blob/ebf327d801610bb3261edd0521f43b7cdf0bc459/services/LinkLoss/source/LinkLossService.cpp#L106)

plantuml/LinkLoss/disconnection.puml
noonfom: Do you want me to remove details about the event queue from the PlantUML?

plantuml/LinkLoss/disconnection.puml
noonfom: That's __not__ true. That event handler must be called here to trigger the event in the application.   

It may be worth duplicating the call in the app column, but removing it completely from the LLS column makes no sense to me.

plantuml/LinkLoss/init.puml
noonfom: This crossed my mind too whilst writing the PR.

noonfom:
@pan- @paul-szczepanek-arm Thanks for the reviews. I see that I have muddled up the terms sequence and activity when referring to the UML diagrams. These are indeed activity diagrams. However, I do plan on adding several success/failure sequence diagrams to the PR to aid in the development of the unit tests.

plantuml/LinkLoss/connection.puml
paul-szczepanek-arm: OK, I was thinking in terms or roles and actions, this is following function implementations

plantuml/LinkLoss/data_written.puml
paul-szczepanek-arm: how is the line 106 interacting with the user application?

plantuml/LinkLoss/disconnection.puml
paul-szczepanek-arm: OK, normally you would put it in between the roles with an arrow pointing the in the direction of the roles being acted on, like this:
```
user              linkloss service              ble                    peer
  +                      +                       +                       +
  |  set alert callback  |                       |                   user sets remote
  +--------------------->+                       |      att write    LLS alert to high
  |                      |                       +<----------------------+
  |                      |       write callback  |                       |
  |                      +<----------------------+                       |
  |               set alert to high              |                       |
  |                      |                       |                     no|tx
  |                      |                    timeout                    |
  |                      |      on disconnect    |                       |
  |                      +<----------------------+                       |
  |                      |                       |                       |
  | call alert callback  |                       |                       |
  +<---------------------+                       |                       |
  |                      |                       |                       |
  +                      |                       |                       |
```

noonfom:
This [force-push](https://github.com/ARMmbed/mbed-os-experimental-ble-services/compare/ebf327d801610bb3261edd0521f43b7cdf0bc459..0e5a2ca4ae78314b61fb3db432394ef79b5ee8be) rebases with master and resolves the confusion highlighted [here](https://github.com/ARMmbed/mbed-os-experimental-ble-services/pull/27#issuecomment-772063594).

plantuml/LinkLoss/connection_normal.puml
```diff
@@ -0,0 +1,17 @@
+@startuml
+participant LinkLossService as LLS
+participant Gap as GAP
+participant ConnectionCompleteEvent as CCE
+
+== Link Loss Service Constructed ==
+LLS <- GAP : onConnectionComplete(event)
+activate LLS
+
+LLS -> CCE : event.getStatus()
```
pan-, line 10: We probably don't need to model the connection completion event, a note explaining what's inside should be enough. There will be another sequence diagram for when the status has an error.

plantuml/LinkLoss/data_written_normal.puml
```diff
@@ -0,0 +1,18 @@
+@startuml
+participant LinkLossService as LLS
+participant Gap as GAP
```
pan-, line 3: I would model the emitter of the `onDataWritten` call as the BLE peer or the BT stack.

plantuml/LinkLoss/data_written_normal.puml
```diff
@@ -0,0 +1,18 @@
...
+participant GattWriteAuthCallbackParams as GWACP
+
+== Link Loss Service Constructed ==
+
+LLS <- GAP : onDataWritten(write_request)
+activate LLS
+
+LLS -> GWACP : level = write_request->data;
```
pan-, line 11: We probably don't need that level of details with the GattWriteAuthCallback param. 
If there's one internal process we want to model it is the data validation. No need to be explicit about that, another diagram can be made when the data in input are not valid.

plantuml/LinkLoss/data_written_normal.puml
```diff
@@ -0,0 +1,18 @@
...
+activate GWACP
+deactivate GWACP
+
+LLS -> LLS : set_alert_level(level)
```
pan-, line 15: This can be replace with `set_alert_level(write_request->data)`

plantuml/LinkLoss/data_written_normal.puml
```diff
@@ -0,0 +1,18 @@
...
+deactivate LLS
```
pan-, line 16: We probably return the status of the authorization to the caller.

pan- (changes requested):
Thanks for the submission @noonfom , there is a high level of detail in your diagram. 
Can you move them into the link loss service and reference the PNG in a `.md` file so they can be easily consulted by anyone looking at the repository in a browser. 

It would be a great addition to add a title for each of those diagrams too.

noonfom:
@pan- I applied your review. Please re review.

pan- (approved):
Awesome, I merge.

PR #32, opened 2021-02-04
Github CI

paul-szczepanek-arm:
Please make PRs of unittests against this branch.

Do not merge - ~this will need rework when Martin makes changes to mbed-os.~ It's ready to go, just waiting for https://github.com/ARMmbed/mbed-os/pull/14426 to l and


tests/unittests.sh
```diff
@@ -0,0 +1,19 @@
...
+cd "$( dirname "$0" )"
+
+# we need stubs from mbed-os
+if [ -d "mbed-os" ]
+then
+    echo "Using existing mbed-os"
+else
+    #git clone https://github.com/ARMmbed/mbed-os.git
+    # until it's not merged we use my branch
+    git clone --depth 1 https://github.com/paul-szczepanek-arm/mbed-os.git -b cmake-tests
```
pan-, line 14: I'm leaving a comment to make sure we don't forget to update this reference

tests/UNITTESTS/CurrentTime/test_CurrentTime.cpp
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    void TearDown()
+    {
+    }
+
+    BLE* ble;
+};
+
+TEST_F(TestLinkLoss, reset)
```
pan-, line 45: That's an excellent explanation of the process and it could serve as a template for testing. 
Can you extend the fixture so it includes things that are common for BLE testing like the event queue the chainable event handlers ? 

I think the file can be moved some place else and naming of the fixture can reflect this is a template/example.

tests/UNITTESTS/CurrentTime/test_CurrentTime.cpp
```diff
@@ -0,0 +1,67 @@
...
+    EXPECT_CALL(ble::security_manager_mock(), reset());
+
+    /* calls are made on real APIs but will fulfill expectations on the mocks as that's what they call in the end */
+    gap.reset();
+    client.reset();
+    server.reset();
+    sm.reset();
+
+    /* remember you must call this at the end of the test if you have any expectations set */
+    ble::reset_mocks();
```
pan-, line 66: This can be part of the tear down.

.gitignore
```diff
@@ -0,0 +1,43 @@
+# Compiled Object files
```
pan-, line 1: Thanks for this.

tests/CI/run-unit-test/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM mbedos/mbed-os-env:latest
```
pan-, line 1: Do we even need a docker file given we just use mbed-os one and branch into the test script ? 
How different is it from invoking `tests/unittests.sh` in the workflow file ?

.github/workflows/main.yml
```diff
@@ -0,0 +1,13 @@
...
+    branches:
+    - main
+  pull_request:
+jobs:
+  build:
+    name: Unit tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ./tests/CI/run-unit-test
```
pan-, line 13: That would be great to add a name on each of these steps. 
Here's an example: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idsteps .

.github/workflows/main.yml
```diff
@@ -0,0 +1,13 @@
+name: CI
```
pan-, line 1: We could probably use a more descriptive name.

.github/workflows/main.yml
pan-, line 13: I also wonder if everything should be driven from a single workflow files or from multiples. As we add more things to be run, it might became harder to read the output and it might take a longer time to execute. 

That's probably something to revisit latter. Not in this PR.

pan- (changes requested):
That's excellent work Paul. Thank you.
I have few improvement request for the CI step: 
- add a run of the tests with valgrind 
- add a run of the tests code coverage and archive the HTML generated
Besides this, I think what's inside the `UNITTESTS/CurrentTime` folder can be moved in another folder so we can keep it as a general template that can be reused by everyone. 

Another point I'd like to discuss, how does it integrate with development of service files ? When a new file is added, the CMakeLists.txt of the service should be modified and the same applies to the CMakeLists.txt of the associated unit tests. Is my understanding correct ? Is it a friction during development ?

tests/UNITTESTS/CurrentTime/test_CurrentTime.cpp
paul-szczepanek-arm: moved

tests/UNITTESTS/CurrentTime/test_CurrentTime.cpp
paul-szczepanek-arm: chainable event handler is not a fake or mock, you use it same way, I don't think we need to duplicate its docs

tests/CI/run-unit-test/Dockerfile
paul-szczepanek-arm: Depends on whether or not you're going to expand on the docker image. I'd keep it to make it consistent - we'll use it for other CI since we need mbed tools there and it has a certain cmake version etc. don't see why we can't run it in a docker to make our life simpler

.github/workflows/main.yml
paul-szczepanek-arm: I left it vague as it's the only one so far and I don't know what else we'll put in there apart from unit tests but I'll rename it for you

.github/workflows/main.yml
paul-szczepanek-arm: I added names

paul-szczepanek-arm:
wait, haven't added valgrind and gcov yet

tests/UNITTESTS/Template/test_Template.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include "ble_mocks.h"
+/* this is a fake event queue that avoids mbed-os dependencies and allows manual dispatch */
+#include "events/EventQueue.h"
+
+using namespace std::chrono_literals;
+
+/* This test does not test anything, you may use it as a template for your unit tests.
+ * It shows all the elements you need to use mocks for all the ble APIS. */
+
+class TestLinkLoss : public testing::Test {
```
pan-, line 37: Can you change the name of the fixture ?

paul-szczepanek-arm:
I need to fix the init mocks thing, give me sec

paul-szczepanek-arm:
beep boop all systems go

scripts/bootstrap.sh
```diff
@@ -0,0 +1,67 @@
...
+source scripts/symlink.sh
+
+# Clone mbed-os
+if [ -d "dependencies/mbed-os" ]
+then
+    echo "Using existing mbed-os"
+else
+    # git clone https://github.com/ARMmbed/mbed-os.git
+    # Use feature branch until merged to master
+    git clone --depth 1 https://github.com/ARMmbed/mbed-os.git -b feature_unittest_refactor dependencies/mbed-os
```
paul-szczepanek-arm, line 29:
```suggestion
    git clone --depth 1 https://github.com/paul-szczepanek-arm/mbed-os.git -b unittest dependencies/mbed-os
```

PR #34, opened 2021-02-04
add docs for running unittests

paul-szczepanek-arm:


tests/README.md
```diff
@@ -0,0 +1,128 @@
+# Testing experimental services
+
+The services provided services are covered by unittests. We use GoogleTest as the testing framework and CTest
```
noonfom, line 3:
```suggestion
The services provided are covered by unittests. We use GoogleTest as the testing framework and CTest
```

tests/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+This will configure and build the tests with Cmake and run the tests.
+
+Unit tests require mbed-os to provide stubs and mocks. A copy of mbed-os will be checked out during the build into
+the tests directory (unless it's already there, you may use a symlink if you already have a checkout of mbed-os).
+This will not be deleted after the run.
+
+# Adding a new test
+
+Before or during the development of your service you should add a new unittest to to cover it. This is done by
```
noonfom, line 26:
```suggestion
Before or during the development of your service you should add a new unittest to cover it. This is done by
```

tests/README.md
```diff
@@ -0,0 +1,128 @@
...
+`mbed-os/UNITTESTS/fakes`.
+
+The most common ones you might need are `mbed-os-fakes-ble` and `mbed-os-fakes-event-queue`.
+
+### mbed-os-fakes-ble
+
+This library provides a fake BLE implementation that uses mocks instead of real BLE components for `Gap`,
+`GattServer`, `GattClient`, `SecurityManager`.
+
+There is no need to initialise the instance, it is ready to go and can get used just like normal BLE instance:
```
noonfom, line 87:
```suggestion
There is no need to initialise a fake BLE instance; it is ready to go and can be used just like a normal BLE instance:
```

tests/README.md
```diff
@@ -0,0 +1,128 @@
...
+The way google test works means that if you set any expectations on your mocks they must be destroyed at
+the end of each test. This is done through the fake BLE instance special method:
+
+```
+ble::reset_mocks();
+```
+
+### mbed-os-fakes-event-queue
+
+This is a fake event queue that doesn't bring in any dependencies from mbed-os. It's API is simplified
```
noonfom, line 118:
```suggestion
This is a fake event queue that doesn't bring in any dependencies from mbed-os. Its API is simplified
```

tests/README.md
```diff
@@ -0,0 +1,128 @@
...
+and it only offers limited functionality.
+
+If you choose to use it you must not also include a library that brings in real headers for the event
+queue as they would conflict.
```
noonfom, line 122: Is this true. Did you not include `events/EventQueue` in your illustrative example?

tests/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+The API calls supported are for simple calls `call`, `call_in` and the `cancel` method.
+
+The event queue is not run in real time and must be progressed manually. You may use
+`dispatch(int milliseconds)` and `dispatch_forever()` to process events in the queue. This way you can
+simulate passage of time within your test.
```
noonfom, line 128:
```suggestion
simulate the passage of time in your test. 
```

noonfom (changes requested):
Thanks @paul-szczepanek-arm. This document is very useful.

tests/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+```
+./unittests.sh
+```
+
+This will configure and build the tests with Cmake and run the tests.
+
+Unit tests require mbed-os to provide stubs and mocks. A copy of mbed-os will be checked out during the build into
+the tests directory (unless it's already there, you may use a symlink if you already have a checkout of mbed-os).
+This will not be deleted after the run.
```
noonfom, line 22: This _might_ be a source of confusion for people. Perhaps you could add the folder structure:
```
tests/
└── CI/
└── UNITESTS/
└── mbed-os/
.mbedignore
CMakeLists.txt
googletest.in
unittests.sh
```

tests/README.md
paul-szczepanek-arm: it is true. that header is a fake header

tests/README.md
paul-szczepanek-arm: OK, I expanded that paragraph.

tests/README.md
noonfom: What's the full path of the fake header in your [cmake_lists](https://github.com/paul-szczepanek-arm/mbed-os/tree/cmake-tests) branch in mbed-os?

tests/README.md
noonfom: Resolved in 54b5a6f18b17b62799b6973382344e625814ded3

tests/README.md
paul-szczepanek-arm: I don't understand the question, look inside UNITTESTS/fakes/ that's where the library is, it has it's own header

tests/README.md
```diff
@@ -0,0 +1,136 @@
+# Testing experimental services
+
+The services provided are covered by unittests. We use GoogleTest as the testing framework and CTest
+as the runner.
+
+When adding a new service please include a unittest suite for your new service in your service PR.
+You are free to use any framework as long as it integrates with our CTest runner but to take advantage of
```
pan-, line 7: Great point.

tests/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+But these are instead google mocks. This means you can set expectations on them. 
+
+```
+EXPECT_CALL(ble::gap_mock(), reset());
+```
+
+This will set up an expectations that at some point during the test the Gap::reset method will be called.
+
+The way google test works means that if you set any expectations on your mocks they must be destroyed at
```
pan-, line 117: Expectations are destroyed when a Mock is destroyed, in the case of ble a singleton is used which makes things more difficult. I would advise on having `ble::reset_mocks` in the teardown of the fixture rather than at the end of each test.

tests/README.md
pan-: I think this line should be clarified. As a user, I would like to know if I can include the event header in my service with a `events/EventQueue.h` and which API is not available in the fake. 

Also what would be the issue if another library includes `events/EventQueue.h` ? It is not possible to propagate the fake event include path when its builded ? That would be a minor issue if it was the case as we try to avoid real dependency in unit tests.

pan- (approved):
Really good, now I want to start and add unit tests :D.

pan- (approved):

PR #43, opened 2021-03-15
rebase github ci branch onto the new unit test feature branch

paul-szczepanek-arm:
I can no longer push against upstream so I made this PR to rebase onto the new mbed-os unittests

PR #44, opened 2021-03-16
Add device information

paul-szczepanek-arm:
Add device information service to replace the obsolete one in mbed-os that will soon be purged.

This is waiting on the https://github.com/ARMmbed/mbed-os-experimental-ble-services/pull/43 to go in.


services/DeviceInformation/include/ble-service-device-information/DeviceInformationService.h
```diff
@@ -0,0 +1,92 @@
...
+#include "ble/BLE.h"
+
+#if BLE_FEATURE_GATT_SERVER
+
+/** BLE Device Information Service class doesn't require instantiation - the characteristics added are read only.
+ * Use the static method add_service to add the chosen Device Information Service characteristics to the server. */
+class DeviceInformationService {
+public:
+    /** System ID */
+    struct SystemId {
```
pan-, line 30: The convention in ble as been to use `snake_case_t` for pods see: https://os.mbed.com/docs/mbed-os/v6.9/contributing/style.html.

services/DeviceInformation/include/ble-service-device-information/DeviceInformationService.h
```diff
@@ -0,0 +1,92 @@
...
+ */
+
+#ifndef BLE_DEVICE_INFORMATION_SERVICE_H
+#define BLE_DEVICE_INFORMATION_SERVICE_H
+
+#include "ble/BLE.h"
+
+#if BLE_FEATURE_GATT_SERVER
+
+/** BLE Device Information Service class doesn't require instantiation - the characteristics added are read only.
```
pan-, line 25: I'm not sure I understand the notice, does it means that characteristics values are immutable on the server side ? 
If the service is immutable we can probably `delete` the constructor and destructor as well as the copy constructor and asignment operators. 
I'd also expect some background explaining what's the service is for.

services/DeviceInformation/source/DeviceInformationService.cpp
```diff
@@ -0,0 +1,127 @@
...
+        param_string_uuids[param_index++] = GattCharacteristic::UUID_SOFTWARE_REVISION_STRING_CHAR;
+    }
+
+    GattCharacteristic* param_chars[9];
+
+    for (int i = 0; i < param_index; i++) {
+        param_chars[i] = new GattCharacteristic(
+            param_string_uuids[i],
+            (uint8_t *)param_strings[i],
+            strlen(param_strings[i]), /* Min length */
```
pan-, line 71: Is there a maximal lenght defined in the specification ?

services/DeviceInformation/source/DeviceInformationService.cpp
```diff
@@ -0,0 +1,127 @@
...
+    if (firmware_revision) {
+        param_strings[param_index] = firmware_revision;
+        param_string_uuids[param_index++] = GattCharacteristic::UUID_FIRMWARE_REVISION_STRING_CHAR;
+    }
+    if (software_revision) {
+        param_strings[param_index] = software_revision;
+        param_string_uuids[param_index++] = GattCharacteristic::UUID_SOFTWARE_REVISION_STRING_CHAR;
+    }
+
+    GattCharacteristic* param_chars[9];
```
pan-, line 65: Do we need an array of pointers to the characteristics. Can we avoid the dynamic allocation ?

pan- (changes requested):
Thanks Paul, that looks like a correct replacement to the service in Mbed OS. I think more detail explaining the purpose of the service would be welcome. It should also mention that the service must be instantiated once.

Seing all the numerc fields, we probably need routines to make sure these numbers are transmitted in the right endianness. 

Last request, can you add an integration test that connects to a device that instantiate the DeviceInformationService. The test would verify the value of the fields.

services/DeviceInformation/source/DeviceInformationService.cpp
paul-szczepanek-arm: `Do we need an array of pointers to the characteristics.`
Yes, that's how service is constructed.
`Can we avoid the dynamic allocation ?`
No, I can allocate on stack or statically, which one do you prefer?

services/DeviceInformation/source/DeviceInformationService.cpp
paul-szczepanek-arm: no

services/DeviceInformation/include/ble-service-device-information/DeviceInformationService.h
paul-szczepanek-arm: Thanks, I'll rename them.

In my defence in BLE we also use CamelCase_t and CamelCase for pods.

services/DeviceInformation/include/ble-service-device-information/DeviceInformationService.h
paul-szczepanek-arm: `I'm not sure I understand the notice, does it means that characteristics values are immutable on the server side ?`
Yes, the characteristic is written only once, presumably at boot. Information here can only change between boots.

I'll remove the constructor.

services/DeviceInformation/include/ble-service-device-information/DeviceInformationService.h
pan-: I think all the new types we defined in the recent year use `snake_case_t` (I might be wrong). It would be nice to have some alignment with general Mbed this time around. 

It think we should start a discussion on this item, Ble not being compliant with other Mbed OS guidelines...

services/DeviceInformation/source/DeviceInformationService.cpp
pan-: It make sense to use the stack when we can. If it's under 500bytes we we can allocate it on the stack.

paul-szczepanek-arm: I don't know how big the characteristics are as they are dynamic so I don't know I'm going to blow your stack. Also if I use the stack I have to allocate all the potential characteristics even if you use one of them. I really don't see the point, this is gatt we're talking about here, dynamic allocation will be made even if I allocate on the stack since each characteristic will malloc the buffer for the data.

pan-: Ok, no strong opinion here, we can keep it that way.

services/DeviceInformation/include/ble-service-device-information/DeviceInformationService.h
```diff
@@ -0,0 +1,110 @@
...
+    /** IEEE 11073-20601 Regulatory Certification Data List */
+    struct regulatory_cert_data_list_t {
+        /** Regulatory and certification information for the product in a list defined in IEEE 11073-20601.
+         *
+         * Must conform to IEEE Std 11073-20601-2008 Health Informatics - Personal Health Device Communication
+         * Application Profile - Optimized Exchange Protocol version 1.0 or later.
+         *
+         * @Attention Pointer must be valid and first byte must be equal to the size of the data
+         * (one byte less than size of the buffer). */
+        uint8_t *data;
```
pan-, line 70: I think we should make it const otherwise it will be difficult to link with variable in the text segment: 

```c++
static constexpr const uint8_t reg_cert_data_list[] = { 
    2, 0xDE, 0xAD
};

void foo() { 
    regulatory_cert_data_list_t reg_cert;
    reg_cert.data = reg_cert_data_list;    /// Compilation error if data is not const!
}
```

pan- (changes requested):

services/DeviceInformation/include/ble-service-device-information/DeviceInformationService.h
paul-szczepanek-arm, line 70:
```suggestion
        const uint8_t *data;
```

pan- (approved):

paul-szczepanek-arm:
not yet, needs a cast because gatt server doesn't like cosnt

services/DeviceInformation/source/DeviceInformationService.cpp
```diff
@@ -0,0 +1,139 @@
...
+            8, /* Min length */
+            8, /* Max length */
+            GattCharacteristic::BLE_GATT_CHAR_PROPERTIES_READ
+        );
+    }
+
+    if (cert_data_list && cert_data_list->data) {
+        param_chars[param_index++] = new GattCharacteristic(
+            GattCharacteristic::UUID_IEEE_REGULATORY_CERTIFICATION_DATA_LIST_CHAR,
+            cert_data_list->data,
```
paul-szczepanek-arm, line 100:
```suggestion
            (uint8_t*)cert_data_list->data,
```

PR #52, opened 2021-03-29
Fix Windows symbolic links

noonfom:
Prepending the `ROOT` variable to relative paths produces incorrect absolute paths on Windows:

```shell
The system cannot find the path specified.
```

`cd` into the root repository and use relative paths instead.

### DIFF
--- a/scripts/activate.sh
+++ b/scripts/activate.sh
@@ -13,8 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Repository root
-ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"/..
+# Enter repository root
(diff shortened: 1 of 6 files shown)


scripts/activate.sh
```diff
@@ -13,8 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Repository root
-ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"/..
+# Enter repository root
+cd "$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"/..
```
paul-szczepanek-arm, line 17:
```suggestion
cd "$( dirname "${BASH_SOURCE[0]}" )"/..
```

paul-szczepanek-arm (approved):

Issue #46, opened 2021-03-17
Service Proposal: FOTAService 


AGlass0fMilk:
### BLE Standard Service:

- [ ] Yes
- [X] No

### Proposed UUID(s) or link to specification document:

See specification and documentation on fota-service branch here:

https://github.com/ARMmbed/mbed-os-experimental-ble-services/tree/5328cacea69452901a691b7380366b6a188cf503/services/FOTA

@pan- @noonfom @paul-szczepanek-arm I took aspects from our discussion in #13 and wrote up a more formal specification README, taking the LinkLossService as an example.

The newly-renamed `FOTAService` strips out some complexity from my original DFUService proposal. I think the new specification covers the most basic use cases while providing for more advanced features and extensions in the future. 

I have yet to refactor the code, but I'm planning to split off the `BlockDevice` dependency into standard `FOTAService::EventHandlers` that we provide. A user can implement their own custom logic if they wish, but we can provide a basic event handler for writing to a given block device. This also serves as a way for us to provide modular extensions (eg: delta updates, trigger connection parameter updates, etc) in the future as part of our standard library of event handlers.

Before I go much further on the code and tests, I wanted to get your feedback.

Let me know what you think and if I missed any corner cases.

pan-:
Thanks for the comprehensive description. It looks **excellent** and it is **easy** to understand how the software would work thanks to the diagrams you provided. 

I'm afraid of an overflow of the sequential fragment ID. Maybe we can pack fragments into _block_ (not in the sense of BockDevice block) or packet of data  that should not exceed a size of `256 * MTU -1`. A new opcode can be introduced to start the packet transaction; it includes the packet size. A new status can be added to report that a packet has been successfully received. 

To extend on that idea, it should be possible for a server to indicate to the client the recommended block size to avoid `XOFF` due to internal processing. 

When `XOFF` is sent by the server, the client may have queued several fragment. In such case it won't know which block must be retransmitted. The fragment ID would be a valuable information to communicate when the server resume transfer with `XON`. In general, `XOFF` and `XON` just like `SYNC_LOST` could include the expected fragment ID byte. 

 One last thing: it would be useful to report to the client the version of the update service so it can adapt if something has to be changed in the flow.

AGlass0fMilk:
> Thanks for the comprehensive description. It looks **excellent** and it is **easy** to understand how the software would work thanks to the diagrams you provided.

Thanks :smile: 

[...]

PR #56, opened 2021-03-31
Add target_compile_definitions to unit testing template

noonfom:
A good point was raised [here](https://github.com/ARMmbed/mbed-os-experimental-ble-services/issues/46#issuecomment-805834211) about the use of configuration parameters in unit tests. If your service defines configuration parameters in an mbed_lib.json file, you must also define them in your CMakeLists.txt, otherwise the build will fail.

This PR adds a [target_compile_definitions](https://cmake.org/cmake/help/latest/command/target_compile_definitions.html) command to the template and updates the unit testing guide.

### DIFF
--- a/tests/UNITTESTS/README.md
+++ b/tests/UNITTESTS/README.md
@@ -53,10 +53,11 @@ target_link_libraries(${TEST_NAME}
 )
 
 target_compile_definitions(${TEST_NAME}
-     PUBLIC
-        # CMake does not generate configuration options;
-        # if your service has an mbed_lib.json, you must define the configuration parameters here
-        BLE_FEATURE_GATT_SERVER=1
+    PUBLIC
+        # CMake does not generate config options;
+        # if your service defines config parameters in an mbed_lib.json,
+        # you must also define them here
+        CONFIGURATION_MACRO_NAME=1
 )
 
 add_test(NAME "${TEST_NAME}" COMMAND ${TEST_NAME})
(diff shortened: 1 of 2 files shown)


tests/UNITTESTS/README.md
```diff
@@ -52,6 +52,13 @@ target_link_libraries(${TEST_NAME}
         gmock_main
 )
 
+target_compile_definitions(${TEST_NAME}
+     PUBLIC
+        # CMake does not generate configuration options;
+        # if your service has an mbed_lib.json, you must define the configuration parameters here
+        BLE_FEATURE_GATT_SERVER=1
```
paul-szczepanek-arm, line 59: for unit tests all ble features are alreayd enabled

paul-szczepanek-arm: https://github.com/ARMmbed/mbed-os/blob/master/connectivity/FEATURE_BLE/include/ble/common/BLERoles.h#L88

paul-szczepanek-arm (approved):
this is very good, we just need a better example because this gives the wrong impression - you don't need to enabled features

paul-szczepanek-arm (changes requested):
maybe CONFIGURATION_MACRO_NAME=1 ?

tests/UNITTESTS/README.md
noonfom: I swapped it out for your suggestion: ffcb41c3a483e79e52d09ebe6eb46f19705ddeeb.

pan- (approved):

PR #60, opened 2021-06-03
FOTA Service

### LINKED ISSUE #46: Service Proposal: FOTAService

AGlass0fMilk:
This PR contains the FOTA service and associated tests as outlined in #46.

Closes #46


AGlass0fMilk:
After working with this service on some projects, I think we should introduce another standard FOTA op-code, "FOTA confirm".

Bootloaders like Mcuboot support rolling back firmware updates if they do not mark themselves as "good". ie: the firmware has a severe bug that prevents normal operation. When reset, mcuboot will then revert the update to prevent "bricking".

One possible failure of this system is, since the update mechanism is in the application, there is a possibility a fatal bug could be introduced into the update mechanism code. This would prevent any future firmware upgrades to the device. The FOTA Confirm op-code would allow the FOTA client to mark the updated firmware as "OK".

This is essentially a self-test of the firmware update mechanism in the BLE FOTA service case -- if BLE can be brought up, a connection can be made, and a characteristic can be written, it is pretty safe to assume that the FOTA code has not been broken by the update.

There are some other recommended self-tests that should be performed, such as testing update flash memory initialization.

I think this would be a common enough use case to deserve a dedicated, standard op-code.

@pan- Thoughts on this? This is a problem for all field-upgradeable devices where the update mechanism is in the application itself. If Mbed-OS creates a framework for DFU, it should include features like a flexible, easy-to-integrate "factory reset" that can reinstall an image stored on a dedicated "recovery partition" (eg: a well-tested application that does nothing but let the firmware be reprogrammed in the case of a broken update).

.github/workflows/run_tests.yml
```diff
@@ -56,3 +56,4 @@ jobs:
           ./tests/TESTS/build.sh -s DeviceInformation -t GCC_ARM -m NRF52840_DK
           ./tests/TESTS/build.sh -s LinkLoss -t GCC_ARM -m NRF52840_DK
           ./tests/TESTS/build.sh -s LinkLoss -t GCC_ARM -m DISCO_L496AG
+          ./tests/TESTS/build.sh -s FOTA -t GCC_ARM -m NRF52840_DK
```
noonfom, line 59: It should also work on disco. There are MCUboot problems but the [mock](https://github.com/noonfom/mbed-os-example-ble/tree/mcuboot/BLE_GattServer_FOTAService/Mock) example works fine locally, i.e. I can transfer a binary into the external flash of my [B-L4S5I-IOT01A](https://os.mbed.com/platforms/B-L4S5I-IOT01A/) using the FOTA service.

**Note:** The disco build is also missing for the Device Information service: #62

common/inc/BaseService.h
```diff
@@ -0,0 +1,40 @@
...
+//
+//    BaseService();
+//
+//
+//
+//
+//};
+
+
+#endif /* MBED_OS_EXPERIMENTAL_BLE_SERVICES_INC_BASESERVICE_H_ */
```
noonfom, line 40: It's a good idea but perhaps we should add it in a separate PR with the other utility files:

* ConnectionThingy.h
* GattPresentationFormatDescriptor.h
* CharacteristicPresentationFormatDescriptor.h
* CharacteristicUserDescriptionDescriptor.h

common/inc/ConnectionThingy.h
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Connection... Thingy?
+ * A common problem encountered when developing BLE applications is:
+ *
+ * Characteristic read/write/notify/indicate events are handled in batch.
+ *
+ * This creates a lot of boilerplate code where a service must filter out by comparing
+ * which handle was written to the callback's parameters, figure out which connection,
+ * and then perform the appropriate action.
```
noonfom, line 31: Do [`setWriteAuthorizationCallback`](https://github.com/ARMmbed/mbed-os/blob/aeaac0e70c76fbc5b753ee8c836e369edf1a7fb4/connectivity/FEATURE_BLE/include/ble/gatt/GattCharacteristic.h#L1556) and [`setReadAuthorizationCallback`](https://github.com/ARMmbed/mbed-os/blob/aeaac0e70c76fbc5b753ee8c836e369edf1a7fb4/connectivity/FEATURE_BLE/include/ble/gatt/GattCharacteristic.h#L1603) not solve this problem?

common/inc/ConnectionThingy.h
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * Maybe connection oriented service? Service logic
+ *
+ *
+ */
+
+
+
+
+#endif /* MBED_OS_EXPERIMENTAL_BLE_SERVICES_COMMON_INC_CONNECTIONTHINGY_H_ */
```
noonfom, line 61: See comment under BaseService.h

common/inc/GattPresentationFormatDescriptor.h
```diff
@@ -0,0 +1,70 @@
...
+	private:
+
+		// Wouldn't it be nice if packing structs was more consistently supported by compilers?
+		//struct GattCharacteristic::PresentationFormat_t format;
+
+		// In lieu of using the struct above, packing issues makes us have to use a raw buffer
+		uint8_t format[7];
+};
+
+#endif /* EP_CORE_FEATURES_FEATURE_BLE_GATTPRESENTATIONFORMATDESCRIPTOR_H_ */
```
noonfom, line 70: See comment under BaseService.h

descriptors/CharacteristicPresentationFormatDescriptor.h
```diff
@@ -0,0 +1,94 @@
...
+    private:
+
+        // Wouldn't it be nice if packing structs was more consistently supported by compilers?
+        //struct GattCharacteristic::PresentationFormat_t format;
+
+        // In lieu of using the struct above, packing issues makes us have to use a raw buffer
+        uint8_t format[7];
+};
+
+#endif /* MBED_OS_EXPERIMENTAL_BLE_SERVICES_DESCRIPTORS_CHARACTERISTICPRESENTATIONFORMATDESCRIPTOR_H_ */
```
noonfom, line 94: See comment under BaseService.h

descriptors/CharacteristicUserDescriptionDescriptor.h
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+private:
+
+    const char* _user_description;
+
+
+};
+
+#endif /* MBED_OS_EXPERIMENTAL_BLE_SERVICES_DESCRIPTORS_CHARACTERISTICUSERDESCRIPTIONDESCRIPTOR_H_ */
```
noonfom, line 59: See comment under BaseService.h

scripts/bootstrap.sh
```diff
@@ -13,59 +13,56 @@
...
 
 # Add symlinks
-symlink dependencies/mbed-os       tests/UNITTESTS/mbed-os
-
-symlink dependencies/mbed-os       tests/TESTS/LinkLoss/device/mbed-os
-symlink services/LinkLoss          tests/TESTS/LinkLoss/device/LinkLoss
+symlink "$ROOT"/dependencies/mbed-os "$ROOT"/tests/UNITTESTS/mbed-os
 
-symlink dependencies/mbed-os       tests/TESTS/DeviceInformation/device/mbed-os
-symlink services/DeviceInformation tests/TESTS/DeviceInformation/device/DeviceInformation
```
noonfom, line 39: Both files are required to build the integration tests for the Device Information service. Can you add them back in?

scripts/bootstrap.sh
```diff
@@ -13,59 +13,56 @@
...
+symlink "$ROOT"/dependencies/mbed-os "$ROOT"/tests/TESTS/FOTA/device/mbed-os
+symlink "$ROOT"/common               "$ROOT"/tests/TESTS/FOTA/device/common
+symlink "$ROOT"/descriptors          "$ROOT"/tests/TESTS/FOTA/device/descriptors
+symlink "$ROOT"/services/FOTA        "$ROOT"/tests/TESTS/FOTA/device/FOTA
 
-# Create mbed-os.lib for CMake builds
-echo "https://github.com/ARMmbed/mbed-os" > tests/TESTS/LinkLoss/device/mbed-os.lib
-echo "https://github.com/ARMmbed/mbed-os" > tests/TESTS/DeviceInformation/device/mbed-os.lib
```
noonfom, line 43: Mbed CLI 2 requires that a `mbed-os.lib` be present in the project root.

The following is also required for the FOTA service:

```shell
echo "https://github.com/ARMmbed/mbed-os" > tests/TESTS/FOTA/device/mbed-os.lib
```

tests/TESTS/FOTA/device/.mbed
```diff
@@ -0,0 +1 @@
+ROOT=.
```
noonfom, line 1: hmm... I don't think we should have .mbed files in device folders.

tests/TESTS/FOTA/device/mbed-os.lib
```diff
@@ -0,0 +1 @@
+https://github.com/ARMmbed/mbed-os
```
noonfom, line 1: That should be added during bootstrap.

common/inc/BaseService.h
AGlass0fMilk: Ha, sorry, these files weren't supposed to be included in this PR :sweat_smile:

common/inc/ConnectionThingy.h
AGlass0fMilk: Unrelated the PR at hand. Those APIs "solve" the problem, but I think it creates a confusing situation in the BLE API:
You can still handle characteristic writes that are authorized but you then have to filter them based on the value handle. It creates two ways to do the same thing which can be confusing to users.

The naming of "setWriteAuthorizationCallback" and "setReadAuthorizationCallback" imply they are _only_ for read authorization/write authorization. Likewise, the naming of `GattServer::EventHandler::onCharacteristicWritten` (or whatever it is) implies it should be where characteristic writes are handled.

PR #67, opened 2021-07-12
add header libs to cmake unittest build

paul-szczepanek-arm:
You have to now add these manually as no longer build by UNITTESTS cmakefile.


pan- (approved):

paul-szczepanek-arm:
wait, need to fix, after more refactoring in mbed-os

paul-szczepanek-arm:
looks like I had some in between version of mbedos - master is fine, re-running to check

